PR #48, opened 2018-02-08
Remove redundant code from AbstractPlannerExecution

dorezyuk:
This pull request cleans up the abstract_planner_execution.cpp by 
1. Uniforming the include style 
2. Removing the checks of has_new_start_ and has_new_goal_ since they are always false at this point
3. Removing the modification of the local variable exceeded since by setting planning_ to false, we exit the while loop


mbf_abstract_nav/src/abstract_planner_execution.cpp
```diff
@@ -354,7 +355,7 @@ namespace mbf_abstract_nav
           ROS_INFO_STREAM("New goal pose: (" << g.x << ", " << g.y << ", " << g.z << ")");
         }
 
-        make_plan = !(success || exceeded) || has_new_start_ || has_new_goal_;
```
spuetz, line 267: We should move this line into the while loop. Somehow it went down to the wrong place, I think.

spuetz: And then maybe remove the `planning_ = false` everywhere. That is much cleaner then. What do you say?

spuetz (changes requested):

mbf_abstract_nav/src/abstract_planner_execution.cpp
corot: These two flags are required. Otherwise, move_base action won't react to new goals

PR #50, opened 2018-02-09
Abstract plugin laoder

dorezyuk:
Pull request contains a version of a common AbstractPluginHandler which hosts the shared code by all Abstract<>Exectution classes. The request contains an implementation for AbstractControllerExecution and might be first merged to a dev branch


corot:
Hi @dorezyuk, thanks a lot for the contribution. I like the code simplification. There's a couple of problems that I have fixed on https://github.com/magazino/move_base_flex/tree/dorezyuk-abstract_plugin_laoder branch (plus some code formatting to keep our standard)

For the missing parts: use for planners and recovery. What's your idea? will you complete it, or should I do it instead (I'm fine with both options)

dorezyuk:
Hi Jorge,
We have considered the move_base_flex library and decided that we cannot use it for out project. Out main objection is the limitation to 2 sequential stages. Instead we have developed a other implementation and are currently evaluating if we can open source it. So it would be better if you complete it.
Best,
Dima

________________________________
Von: Jorge Santos Simón <notifications@github.com>
Gesendet: Sonntag, 18. Februar 2018 22:26
An: magazino/move_base_flex
Cc: Dima Dorezyuk; Mention
Betreff: Re: [magazino/move_base_flex] Abstract plugin laoder (#50)


Hi @dorezyuk<https://github.com/dorezyuk>, I like the code simplification. There's a couple of problems that I have fixed on https://github.com/magazino/move_base_flex/tree/dorezyuk-abstract_plugin_laoder branch (plus some code formatting to keep our standard)

For the missing parts: use for planners and recovery. What's your idea? will you complete it, or should I do it instead (I'm fine with both options)

corot:
Sorry to hear MBF doesn't fit your needs. As we want to make MBF as generally useful as possible, I would you like to have a short talk about what your needs are and what's missing on MBF to address them. Maybe your case is more common that we had expected. I'm also based on Munich, so we can either chat or take a beer togeher.

dorezyuk:
Hi Jorge
Yes we can. We have met already at the last meet up magazino hosted.

Am 20.02.2018 um 11:32 schrieb Jorge Santos Simón <notifications@github.com<mailto:notifications@github.com>>:


Sorry to hear MBF doesn't fit your needs. As we want to make MBF as generally useful as possible, I would you like to have a short talk about what your needs are and what's missing on MBF to address them. Maybe your case is more common that we had expected. I'm also based on Munich, so we can either chat or take a beer togeher.

spuetz:
I thought about how to make the plugin executions of multiple plugins concurrent. So we definitively need a separate abstract plugin factory or loader class which is used by the server and which holds all plugins,  loads and unloads all plugins. The server then hands over the selected plugin instances to the execution class instances. That would be a clean way of integrating concurrent actions. So we need to wait with this PR.

spuetz:
I close this due to the plugin manager and the abstract base execution in #78

PR #61, opened 2018-03-30
Updated move_base_simple/goal to a relative topic

Timple:
The original topic of move_base is relative. This nodes escapes its namespace. So I removed the prepended slash.


spuetz:
You're right! Thx!

PR #67, opened 2018-04-27
fix: Remove `xmlrpcpp` prefix

alireza-hosseini:
# What I did
 - since ROS **kinetic**, The `XmlRpcException.h` has been moved to the **xmlrpcpp** directory,
   while on the **indigo** distro it is located in the `$ROS_DISTRO/include`, accordingly the `#include <xmlrpcpp/XmlRpcException.h>` wouldn't work on this distro.
 - This PR provides the backward compatibility by removing the `xmlrpcpp` prefix so that it can be compiled on both of **kinetic** and **indigo** distros.
 - FYI: As it can be seen in the following link, in the `abstract_recovery_execution.cpp`: https://github.com/magazino/move_base_flex/blob/9c5fb1b47b94a32d575b5544466fb4d15450c8f7/mbf_abstract_nav/src/abstract_recovery_execution.cpp#L41 it is already without`xmlrpcpp` prefix, while on the other two `cpp` files it wasn't.
 
# How I tested
- By compiling the **mbf_abstract_nav** package on both distros

### DIFF
--- a/mbf_abstract_nav/src/abstract_controller_execution.cpp
+++ b/mbf_abstract_nav/src/abstract_controller_execution.cpp
@@ -39,7 +39,7 @@
  */
 
 #include "mbf_abstract_nav/abstract_controller_execution.h"
-#include <xmlrpcpp/XmlRpcException.h>
+#include <XmlRpcException.h>
 #include <mbf_msgs/ExePathResult.h>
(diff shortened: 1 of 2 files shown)


corot:
thanks!

Issue #36, opened 2018-02-02
Allow running more than one planner and recovery behavior simultaneously


corot:
Wouldn't it make sense to allow running more than one planner and recovery behavior simultaneously?
Some use cases:
* Run two complementary recovery behaviors simultaneously so they help each other, or at least reduce the overall time compared to sequential execution
* Run two or more different planners simultaneously and use the first / best plan obtained
* Start simultaneously fast and slow plannings so we can start moving asap but a bit after replace the "fast" plan by a (probably much better) "slow" plan

corot:
Implemented with #78

PR #76, opened 2018-07-06
Use stringstream to append enum to message

ipa-fez:
Adding the int to the string literal does pointer arithmetic.

### DIFF
--- a/mbf_abstract_nav/src/abstract_navigation_server.cpp
+++ b/mbf_abstract_nav/src/abstract_navigation_server.cpp
@@ -39,6 +39,7 @@
  */
 
 #include <nav_msgs/Path.h>
+#include <sstream>
 #include "mbf_abstract_nav/abstract_navigation_server.h"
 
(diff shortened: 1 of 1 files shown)


corot:
lgtm

MichaelGrupp:
Drive-by comment: `std::to_string(int)` if C++11 is used?

corot:
Nop...  as we started development on indigo, we decided not to use C+11.
Probably is time to reconsider that

MichaelGrupp:
Would be cool, 11 is just so much superior IMO.

spuetz:
@MichaelGrupp After #36 merging branch https://github.com/magazino/move_base_flex/tree/feature/concurrency we will switch to c++11 or c++14, since kinetic has the target version c++11 and melodic has c++14. Older versions do not make sens any more. We live in 2018 and the old c++ standard is limiting too much.

spuetz (approved):
@ipa-fez thank you for the pull request!

PR #78, opened 2018-07-18
Concurrency for planners, controllers and recovery behaviors 

spuetz:
Hi all,

so I almost finished the concurrency feature #36  I needed to restructure the code a lot! I wasn't able to do it in smaller pull requests, because the things in the architecture are connected interdependently. The code now is in a much better structure and much more overseeable and clear.

I will explain the pull request in detail soon. I would appreciate if some could test the feature and list the things which are missing, there might be some things. A list of TODOs would be nice.


mbf_abstract_nav/include/mbf_abstract_nav/abstract_action.h
```diff
@@ -0,0 +1,113 @@
...
+    if(elem != executions_.end())
+    {
+      boost::lock_guard<boost::mutex> lock_guard(map_mtx_);
+      elem->second->cancel();
+    }
+  }
+
+  void runAndCleanUp(GoalHandle goal_handle, typename Execution::Ptr execution_ptr){
+    run_(goal_handle, *execution_ptr);
+    ROS_DEBUG_STREAM("Finished action run method, wait for execution thread to finish.");
```
MichaelGrupp: s/wait/waiting ?

mbf_abstract_nav/src/abstract_execution_base.cpp
```diff
@@ -0,0 +1,47 @@
...
+  bool AbstractExecutionBase::start()
+  {
+    //setState(STARTED); // TODO
+    thread_ = boost::thread(&AbstractExecutionBase::run, this);
+    return true;
+  }
+
+  void AbstractExecutionBase::stop()
+  {
+    ROS_WARN_STREAM("Try to stop the planning rigorously by interrupting the thread!");
```
MichaelGrupp: s/Try/Trying ?

MichaelGrupp:
I'm not actively working on MBF, so I just had a quick look for some nits. Added @MirkoFerrati if he has some time.

mbf_abstract_nav/src/move_base_action.cpp
```diff
@@ -0,0 +1,590 @@
...
+     action_client_recovery_(private_nh_, "recovery"),
+     oscillation_timeout_(0),
+     oscillation_distance_(0),
+     recovery_enabled_(true),
+     cancel_(false),
+     exe_path_canceled_(false),
+     behaviors_(behaviors),
+     action_state_(NONE),
+     recovery_trigger_(NONE),
+     replanning_rate_(private_nh_.param<double>("replanning_frequency", 1.0))
```
SHTseng: Some applications may not need the re-planning function. Whereas, ros::Rate does not allow value to be 0, or it will crash in next cycle.

spuetz: Thanks for your comments and the testing! Will check that and find a solution!

mbf_costmap_nav/src/mbf_costmap_nav/costmap_controller_execution.cpp
```diff
@@ -37,95 +37,33 @@
...
-  }
-
-  if (!costmap_ptr_)
-  {
-    ROS_ERROR_STREAM("The costmap pointer has not been initialized!");
-    return false;
-  }
-
   ros::NodeHandle private_nh("~");
   private_nh.param("controller_lock_costmap", lock_costmap_, true);
```
SHTseng: Update the parameter with each cmd_vel cycle seems to be redundant. Maybe these two lines to the constructor?

SHTseng:
I've test it with turtlebot2 using two controllers, trajectory planner and DWA planner. It works fine, although i do not know how the navigation server selects cmd_vels. I will continue test the PR in the next few days.

corot (commented):
Not fully tested, but I saw some initial problems. I'm adding fixes to some on PR78_review branch (those marked in the list):
- [x] global planner not blocking costmap: parameter planner_lock_costmap is ignored
- [x] no replanning on move_base action calls: parameter planner_frequency is ignored
- [x] shutdown_costmaps dyn. reconfigure flag stops costmaps update even when moving (not super-elegant solution, admittedly)
- [x] boost::thread_group threads_ on AbstractAction class grows unbounded (my solution is rather ugly, sorry)
- [x] fix dynamic reconfiguration

mbf_abstract_nav/src/controller_action.cpp
```diff
@@ -0,0 +1,267 @@
...
+    const std::string &action_name,
+    const RobotInformation &robot_info)
+    : AbstractAction(action_name, robot_info, boost::bind(&mbf_abstract_nav::ControllerAction::run, this, _1, _2))
+{
+}
+
+
+void ControllerAction::before(GoalHandle &goal_handle, AbstractControllerExecution& execution)
+{
+  if(threads_.size() == 1)
```
corot: Should not be >= 1?  If not, only first call to controller works, as threads_ keeps increasing size with every call

corot: Should not be >= 1? If not, only first call to controller works, as threads_ keeps increasing size with every call

corot (changes requested):

mbf_abstract_nav/src/controller_action.cpp
```diff
@@ -0,0 +1,267 @@
...
+
+
+void ControllerAction::after(GoalHandle &goal_handle, AbstractControllerExecution& execution)
+{
+  if(goal_handle.getGoalID().id == cmd_vel_pub_goal_id_ && threads_.size() > 1)
+  {
+    execution.disablePublishCmdVel();
+    cmd_vel_pub_goal_id_ = goal_ids_.front();
+    AbstractControllerExecution::Ptr other_exec = executions_[cmd_vel_pub_goal_id_];
+    other_exec->enablePublishCmdVel();
```
corot: I can confirm changing to >= 1 fixes the issue, but then I get a crash on line 35 when sending three consecutive goal w/o waiting for any to finish. Looks like other_exec pointer is not valid

corot: As @spuetz suggested, skipping the last two commits fixes both issues

corot (approved):

PR #83, opened 2018-09-20
Make concurrent move_base_flex melodic compatible

jabbenseth:
The main changes were
1) somehow the signature of actionlib::ActionServer doesn't support goal and cancel callbacks with references anymore (although I have not found out how it worked before)

2) Use tf2 if running on melodic: costmap2d switched to tf2 with melodic. Depending on the ros-distro different typedefs are used and in mbf_utility the pose transformation is done with tf2 for melodic

### DIFF
--- a/mbf_abstract_nav/CMakeLists.txt
+++ b/mbf_abstract_nav/CMakeLists.txt
@@ -50,9 +50,13 @@ catkin_package(
   DEPENDS Boost
 )
 
-if("$ENV{ROS_DISTRO}" STREQUAL "melodic")
-  add_definitions(-DCOSTMAP_HAS_TF2)
-  message("melodic compatible build")
+if("$ENV{ROS_DISTRO}" STREQUAL "kinetic")
+  add_definitions(-DUSE_OLD_TF)
+  message("kinetic compatible build")
+endif()
+if("$ENV{ROS_DISTRO}" STREQUAL "lunar")
+  add_definitions(-DUSE_OLD_TF)
+  message("lunar compatible build")
 endif()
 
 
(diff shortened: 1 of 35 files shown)


corot (approved):
Thanks a lot! I was trying to find some time to install melodic and fix the compilation error on action servers (due to new boost version).

jabbenseth:
I'm not totally sure about the fix though. The passing is now by value, but since it is a handle consisting only of shared pointers and the copy constructor / `GoalHandle::operator=` does not create a deepcopy so it should be fine

mbf_abstract_nav/CMakeLists.txt
```diff
@@ -50,6 +50,12 @@ catkin_package(
   DEPENDS Boost
 )
 
+if("$ENV{ROS_DISTRO}" STREQUAL "melodic")
```
spuetz: We should make it the other way around. So for `kinetic` and `luna` the define `USE_OLD_TF`. New ros distros would be autometically included.

spuetz (changes requested):
Thanks for the pull request! Just one little thing I would chnage. All the rest looks good to me! @corot What do you think?

mbf_abstract_nav/CMakeLists.txt
corot: sounds reasonable, yes.  always looking forward!

For the change on the action servers, remove the reference is the first thing I wanted to try, so... I feel confirmed.

jabbenseth:
not yet tested on kinetic

spuetz:
I'll give it a try ;)

doisyg:
Can anybody confirm  that mbf is working under Melodic?

corot:
Not yet, but I will install Melodic on next week and do so.

PR #84, opened 2018-09-20
navfn indigo compatibility

jabbenseth:
similar pattern as for melodic which allows indigo compatibility


jabbenseth:
I will reopen with a similar pattern and naming scheme as introduced for melodic in #83

PR #92, opened 2018-10-02
Check ROS version on code instead of on CMake

corot:
Otherwise, we force external plugins to do the same. Now MBF and plugins should compile on all ROS versions.

Please @ipa-jba, can you verify that you can still compile on melodic and indigo?

### DIFF
--- a/mbf_costmap_nav/src/nav_core_wrapper/wrapper_global_planner.cpp
+++ b/mbf_costmap_nav/src/nav_core_wrapper/wrapper_global_planner.cpp
@@ -50,11 +50,13 @@ uint32_t WrapperGlobalPlanner::makePlan(const geometry_msgs::PoseStamped &start,
                                         double &cost,
                                         std::string &message)
 {
-#ifdef USE_OLD_NAV_CORE
+#if ROS_VERSION_MINIMUM(1, 12, 0) // if current ros version is >= 1.12.0
+  // Kinetic and beyond
+  bool success = nav_core_plugin_->makePlan(start, goal, plan, cost);
+#else
+  // Indigo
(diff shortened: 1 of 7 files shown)


mbf_costmap_nav/src/nav_core_wrapper/wrapper_global_planner.cpp
```diff
@@ -50,11 +50,13 @@ uint32_t WrapperGlobalPlanner::makePlan(const geometry_msgs::PoseStamped &start,
                                         double &cost,
                                         std::string &message)
 {
-#ifdef USE_OLD_NAV_CORE
+#if ROS_VERSION_MINIMUM(1, 10, 0) // if current ros version is >= 1.10.0
```
jabbenseth: indigo is minor version 11 and jade (which should have been 11) is not even in the list anymore (maybe messed up somehow) so to support the LTS versions this has to be `ROS_VERSION_MINIMUM(1,12,0)`.

jabbenseth (changes requested):
I tested it with industrial-ci (which by the way successfully runs for indigo, kinetic and melodic given the change of the ROS_VERSION) and on my melodic machine.

--> other than the somehow messed up minor versions of ros :+1:

mbf_costmap_nav/src/nav_core_wrapper/wrapper_global_planner.cpp
corot: right! thanks.   thanks for testing!

howardcochran:
I vote for this change. I recently ran into trouble when my plugin would not compile because it had not defined USE_OLD_TF.  This approach is better.

spuetz (approved):

Issue #93, opened 2018-10-04
SIGSEGV crash due to thread unsafe cancel & restart action


howardcochran:
I tracked down a memory corruption bug as being caused by lack of thread safety as noted by this comment in mbf_abstract_nav/.../abstract_action.h:
`if(slot != concurrency_slots_.left.end()) // if there is a plugin running on the same slot, cancel it // TODO make thread safe`

External code was sending a RecoveryAction goal, canceling it, then sending another goal the same behavior name and concurrency slot.  The problem occurs if the recovery plugin takes some time to cancel. Since AbstractAction::start() for the canceled action to finish, it would invoke the same recovery behavior object concurrently even though they are in the same concurrency slot. The two threads operating on the same plugin object trashing each other's memory.

Note that there is no fundamental reason why the same problem could not occur with a GetPath action rather than Recovery. However, it does not happen with ExePath since it has special code to replace the existing ExePath goal rather than canceling and starting the plugin anew.

howardcochran:
### Here is a quick test program that exhibits the behavior when run on a recovery behavior that is non-cancelable or which takes more than a fraction of a second to exit when canceled:

    #!/usr/bin/env python
    
    import sys
    import threading
    import rospy
    import actionlib
    from std_msgs.msg import Header
    import mbf_msgs.msg as mbf_msgs
    
    class RecoverySpammer(object):
        def __init__(self, recovery_name):
            self.recovery_ac = actionlib.SimpleActionClient(
                    '/move_base/recovery', mbf_msgs.RecoveryAction)
            self.recovery_name = recovery_name
            self._is_recovery_running = False
            self._ac_lock = threading.Lock()
            self.timer = rospy.Timer(rospy.Duration(0.5), self._timer_cb)
    
        def start_recovery(self):
            with self._ac_lock:
                rospy.loginfo("Start Recovery")
                goal = mbf_msgs.RecoveryGoal(behavior=self.recovery_name)
                self.recovery_ac.stop_tracking_goal()  # In case old goal exists
                self.recovery_ac.send_goal(goal, done_cb=self._recovery_done_cb)
                self._is_recovery_running = True
    
        def cancel_recovery(self):
            # Note: Not holding self.ac_lock is intentional, as it could deadlock
            rospy.loginfo("Cancel Recovery")
            self.recovery_ac.cancel_goal()
            ## self.recovery_ac.cancel_all_goals()
            with self._ac_lock:
[...]

PR #94, opened 2018-10-04
RFC fix crash thread unsafe cancel action

howardcochran:
This is an "Request for Comments" fix for #93 "SIGSEGV crash do to thread unsafe cancel & restart action" and is not intended to be merged in its current form.  Please read the individual commit comments. It is intended to start discussion on the best fix.


spuetz:
@howardcochran Shouldn't the mutex lock be enough? Why do we need a worker thread for the pending threads? All threads must wait before the lock. Since the method should be executed quite fast. So, I only see the bug of using the lock at a wrong point in the cancel method. Why do we need to wait for the cancel to be finished?

spuetz:
Solved by #138

PR #95, opened 2018-10-16
Do not publish path from MBF

corot:
It's published by all planners, afaik.
I also moved the goal_pose to the private node handle, so it comes within MBF namespace

### DIFF
--- a/mbf_abstract_nav/include/mbf_abstract_nav/planner_action.h
+++ b/mbf_abstract_nav/include/mbf_abstract_nav/planner_action.h
@@ -66,12 +66,6 @@ class PlannerAction : public AbstractAction<mbf_msgs::GetPathAction, AbstractPla
  protected:
 
   /**
-   * @brief Publishes the given path / plan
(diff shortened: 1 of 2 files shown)


spuetz:
The plan is published from every planner, yes, shouldn't we add a switch / param for publishing it or not? I mean this breaks the current API a bit. So we should discuss that first.

corot:
> The plan is published from every planner, yes, shouldn't we add a switch / param for publishing it or not? I mean this breaks the current API a bit. So we should discuss that first.

Nope, path was not published on old move_base, if I checked properly

spuetz:
No I mean API change of MBF. So why not having a switch for that using a param for turning it on and off

corot:
mmm... I don't think anyone is using this path... I even noticed by chance that we where publishing it! :see_no_evil:

spuetz:
I would like to keep the publishPlan method in MBF.

spuetz:
Ok, rethought. I'll merge it.

Issue #100, opened 2018-10-25
Global planner planning from map origin instead of robot position


doisyg:
Hello,
I am new to move base flex (under melodic) and I am trying to migrate from move_base using mbf_costmap_nav. 
I must be missing something (a param somewhere) but when sending a goal to the robot, the global plan is made from the origin of the map instead of the robot position. I tried with both `global_planner/GlobalPlanner` and `carrot_planner/CarrotPlanner`, same behavior. It seems that the global planner can't get the robot position. Where is it suppose to get the robot frame from ? `/mbf_costmap_nav/global_costmap/robot_base_frame` ?

corot:
Interesting...  so you send the goals from RViz? How are you exactly calling MBF?  Thanks

doisyg:
Yes goal are sent from Rviz. 
I tried first with the state machine of the tutorial : http://wiki.ros.org/move_base_flex/Tutorials/Simple%20SMACH%20for%20Move%20Base%20Flex (note: i had to replace `move_base_flex_msgs` by `mbf_msgs` and `move_base_simple` by `mbf_costmap_nav`).
And then with move_base_legacy_relay.py ( again replacing `move_base_flex` by `mbf_costmap_nav`  for the path to the action in the code)
Same result in both case. I believe the controller is fine as after struggling to get to the global path
(although wrong) it could execute it properly.

Below is a screenshot from Rviz where we can see the global path generated by the carrot planner starting at the origin at the map (instead at the position of the robot) and finishing at the goal sent by rviz

![carrot_planner](https://user-images.githubusercontent.com/15727892/47568017-47bc9900-d930-11e8-8bfd-63fb2bcbf0cb.png)

spuetz:
Can you check with which frame the goal is published? Is it published from base_footprint? What is the fixed frame you are using?

doisyg:
The goal is published in the map frame. I use base_link as the fixed frame (it is the same as base_footprint with a Z offset in my config). See below rostopic output.

```
~$ rosparam get /mbf_costmap_nav/global_costmap/global_frame
map
```

```
~$ rosparam get /mbf_costmap_nav/global_costmap/robot_base_frame 
base_link
```

```
~$ rostopic echo /move_base_simple/goal 
header: 
  seq: 1
  stamp: 
    secs: 196
    nsecs: 747000000
  frame_id: "map"
pose: 
  position: 
    x: 10.391412735
    y: 13.6700534821
    z: 0.0
[...]

PR #106, opened 2018-11-09
Fix getRobotPose in melodic

ipa-fez:
The wrong frame was passed to transformPose. Somehow this worked pre-melodic.

### DIFF
--- a/mbf_utility/include/mbf_utility/navigation_utility.h
+++ b/mbf_utility/include/mbf_utility/navigation_utility.h
@@ -66,7 +66,7 @@ namespace mbf_utility
  * @param out Transformed pose.
  * @return true, if the transformation succeeded.
  */
-bool transformPose(const TF &tf_listener,
+bool transformPose(const TF &tf,
                    const std::string &target_frame,
                    const ros::Time &target_time,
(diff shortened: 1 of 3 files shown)


corot:
LGTM; works both on kinetic and melodic, and fixes #100

spuetz (changes requested):
Hmmm, there are some things mixed up. But I'll review the rest tomorrow morning. (In 13 h)

spuetz (changes requested):
We could also improve this by using lookUpTransform in both cases and use the transformation directly by applying it to the pose object, which is a transform object.

spuetz:
@ipa-fez Could you test this? https://github.com/magazino/move_base_flex/tree/fix-melodic-tf I fixed the frame ids of this PR. Thank you!

ipa-fez:
Seems to work fine on melodic @spuetz, thanks for the improvements! I did indeed get fixed frame and target frame mixed up.

ipa-fez:
I pushed the commits to this branch. Feel free to merge.

corot:
hehe  I was just about to merge this

PR #107, opened 2018-11-14
Revert "fix for empty argument"

Rayman:
This reverts commit 2f953dc55feef61d4a730d90d3cc198ab5660db9.

We have been testing move base flex and it crashes everytime when we send a goal when a previous goal is running. Reverting the following commit fixes the issue.

Our test was as follows:
1. run mbf_costmap_nav & mbf_legacy_relay
2. send a goal
3. while the first goal is running, immediately send a new goal
4. when the robot is arriving, the robot will crash with the following segfault:
```
#0  mbf_abstract_nav::MoveBaseAction::actionExePathDone (this=0x6386d8, state=..., result_ptr=...) at /home/ramon/ros/fleet/src/mbf_abstract_nav/src/move_base_action.cpp:368
#1  0x00007ffff5e362e8 in boost::function2<void, actionlib::SimpleClientGoalState const&, boost::shared_ptr<mbf_msgs::ExePathResult_<std::allocator<void> > const> const&>::operator() (a1=..., a0=..., this=0x638c18)
    at /usr/include/boost/function/function_template.hpp:773
#2  actionlib::SimpleActionClient<mbf_msgs::ExePathAction_<std::allocator<void> > >::handleTransition (this=0x638a98, gh=...) at /opt/ros/kinetic/include/actionlib/client/simple_action_client.h:546
#3  0x00007ffff5e0062b in boost::_mfi::mf1<void, actionlib::SimpleActionClient<mbf_msgs::ExePathAction_<std::allocator<void> > >, actionlib::ClientGoalHandle<mbf_msgs::ExePathAction_<std::allocator<void> > > >::operator() (a1=..., p=<optimized out>,
    this=<optimized out>) at /usr/include/boost/bind/mem_fn_template.hpp:165
#4  boost::_bi::list2<boost::_bi::value<actionlib::SimpleActionClient<mbf_msgs::ExePathAction_<std::allocator<void> > >*>, boost::arg<1> >::operator()<boost::_mfi::mf1<void, actionlib::SimpleActionClient<mbf_msgs::ExePathAction_<std::allocator<void> > >, actionlib::ClientGoalHandle<mbf_msgs::ExePathAction_<std::allocator<void> > > >, boost::_bi::list1<actionlib::ClientGoalHandle<mbf_msgs::ExePathAction_<std::allocator<void> > >&> > (a=<synthetic pointer>, f=..., this=<optimized out>) at /usr/include/boost/bind/bind.hpp:313
[...]

PR #108, opened 2018-11-18
Cancel any running action client on receiving a new move_base goal

corot:
This should fix #107 the right way. @Rayman, could you please give a try to this PR ant report if it works?
Thanks a lot


corot:
Nop, my bad. As @Rayman showed, this actually doesn't fix the issue.

Issue #102, opened 2018-10-29
Feature request: Add cmd_vel to ExePathFeedback


reinzor:
For monitoring the robot's intention when a the controller does not return success (due to a collision check for example).

Would you be open for this? I can submit PR :)

corot:
Makes sense; PR always welcome! :neckbeard:

reinzor:
I am working on this but it seems that the `current_twist` is implemented incorrectly.

The `current_twist` is filled by `execution.getLastValidCmdVel()`. I though this would be the actual robot velocity. Any thoughts?

howardcochran:
Hello.  I'd just like to give my two cents here.  I think that the feedback should show the current_twist that is calculated by the controller (i.e. the commanded velocity), not the actual velocity. For anyone who wants to know the actual velocity, you can subscribe to /odom and get it from there. But the ExePathFeedback is giving feedback from the controller so it seems right that it should say what the controller is commanding.

reinzor:
I agree with @howardcochran , maybe it was already intended as this. @corot , could you comment?

corot:
Yes, @howardcochran is right, current_twist provides the last calculated command.
If I understood your intention, @reinzor, you want that this twists gets provided as feedback even in case of failure, right? This means moving [this line](https://github.com/magazino/move_base_flex/blob/master/mbf_abstract_nav/src/abstract_controller_execution.cpp#L338) above the _if (outcome_ < 10)_, but given than in the else we call _publishZeroVelocity_, to be coherent the current_twist feedback should be a zero velocity.
Otherwise you would be "lying".

reinzor:
That is correct. At least it is nice to see the controller's intention when something is blocking the path for example. However, I think the initial issue came from the fact that the naming is a little bit misleading. `current_twist` sound more like the actual `twist` instead of the `twist_reference` or `twist_command` or `cmd_vel`.

corot:
I agree...  specially because it comes together with current_pose, what is actually the current pose of the robot...
I suggest instead last_cmd_vel. @spuetz, stimmt?

reinzor:
Covered by https://github.com/magazino/move_base_flex/pull/120

PR #120, opened 2019-02-27
Issue #102: Change current_twist for last_cmd_vel on exe_path/feedback

corot:
Is a clearer name, as current_twist can be interpreted as the actual velocity of the robot.
Also ensure that we report a zero velocity when we send it to the robot

### DIFF
--- a/mbf_abstract_nav/src/abstract_controller_execution.cpp
+++ b/mbf_abstract_nav/src/abstract_controller_execution.cpp
@@ -192,6 +192,7 @@ namespace mbf_abstract_nav
     vel_cmd_stamped_ = vel_cmd;
     if (vel_cmd_stamped_.header.stamp.isZero())
       vel_cmd_stamped_.header.stamp = ros::Time::now();
+    // TODO what happen with frame id?
   }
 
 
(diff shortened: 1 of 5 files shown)


mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -364,7 +362,12 @@ namespace mbf_abstract_nav
             }
             // could not compute a valid velocity command -> stop moving the robot
             publishZeroVelocity(); // command the robot to stop
+            cmd_vel_stamped.twist = geometry_msgs::Twist();  // overwrite whatever the plugin calculated
```
reinzor: Why not still expose the controller's intention? The controller could say that the path is blocked but still generate a reference to denote its intention?

corot: Well, it would be a lie! last_cmd_vel informs about the last command send to the robot, and in case of error, it will be zero-velocity. Do you think makes more sense to provide the last command generated by the plugin? :thinking: Could be... can you mention a use case when this semantic is useful?

reinzor: The controller's intention can be different from the command sent to the robot. We have a separate collision check module that updates the return value of the controller, e.g. path_blocked or success. However, the control law that the controller is using could be based for example on distance to path and angle to path. So it can still generate a reference. However, the collision check module prevents this reference of being executed.

This way you can highlight a led or speak up if someone is blocking at the front or rear of the robot. Or ask to step aside. At least denote to the used what the robot's intention 

It comes down to semantics of the `last_cmd_vel`, whether it is:
- last command sent to robot
- controller's intention

corot: changed

PR #123, opened 2019-05-08
Fix controller_patience when controller_max_retries is -1

Rayman:
When controller_max_retries is -1, the parameter controller_patience
does not have any effect. The controller is stuck in an infinite loop.

The issue is that setVelocityCmd() keeps setting the
getLastValidCmdVelTime() time to ros::Time::now(). This leads to the
patience code never triggering.

The commit 68664fd7a79d1a7ad035b194ce3e6e1ab7fff8fa has introduced this
bug. This commit moves the two lines back to their original location.

### DIFF
--- a/mbf_abstract_nav/src/abstract_controller_execution.cpp
+++ b/mbf_abstract_nav/src/abstract_controller_execution.cpp
@@ -334,6 +334,9 @@ namespace mbf_abstract_nav
 
           if (outcome_ < 10)
           {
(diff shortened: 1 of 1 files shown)


corot:
Hi, thanks for the fix. You are right that  commit 68664fd broke the controller_patience, but please take a look at the alternative PR #124, where I fix the issue without reverting 68664fd.
Can you please try it?

I also added some comments and renamings, trying to make the code clearer

Rayman:
Yes, that also fixes my issue. Thanks, i'll close this PR then.

PR #124, opened 2019-05-12
Fix controller_patience when controller_max_retries is -1

corot:
Alternative fix to #123 without reverting #102.

I also tried to make clearer the two patience cases and make naming more coherent


Rayman (approved):
I've tested that this PR fixes the issue I had in #123. LGTM.

Issue #121, opened 2019-03-27
Transitioning goal to LOST


Salem-Guezguez:
Hey , I was recently playing around with the paramters of my planners (dwa and eband) looking for a best combination that leads to a good behaviour and I often faced  these two error messages.

[ERROR] [1553594179.855549955, 32.856000000]: NO PATH!
[ERROR] [1553594179.856440514, 32.856000000]: Failed to get a plan from potential when a legal potential was found. This shouldn't happen.
[move_base_flex-4] process has died [pid 15548, exit code -11, cmd /home/salem/luhbots_ws/devel/lib/mbf_costmap_nav/mbf_costmap_nav __name:=move_base_flex __log:=/home/salem/.ros/log/529d7384-4fab-11e9-824f-dc85def86db3/move_base_flex-4.log].

[ INFO] [1553594323.573450136, 162.514000000]: Got new plan
[ INFO] [1553594329.042543536, 167.367000000]: Goal reached
[ WARN] [1553594329.355166851, 167.647000000]: Transitioning goal to LOST
[move_base_flex-4] process has died [pid 18347, exit code -11, cmd /home/salem/luhbots_ws/devel/lib/mbf_costmap_nav/mbf_costmap_nav __name:=move_base_flex __log:=/home/salem/.ros/log/529d7384-4fab-11e9-824f-dc85def86db3/move_base_flex-4.log].
log file: /home/salem/.ros/log/529d7384-4fab-11e9-824f-dc85def86db3/move_base_flex-4*.log

the second error always pops up. what is the source of these error? are they instigated by my changing of one of the parameters, one that i shouldn't have?

praveenramanujam:
@Salem-Guezguez i had a similar issue but when recovery_behavior was added to the ros parameter, we dont face this issue anymore. the cancel goal seems to look for recovery

corot:
The "Failed to get a plan from potential when a legal potential was found. This shouldn't happen." error is caused by [global_planner](http://wiki.ros.org/global_planner). I saw it before, but I have no idea where it comes from. That said, MBF should not crash; I'll investigate it

corot:
Solved by #138

PR #135, opened 2019-06-06
Prevent LOST goals when replanning

corot:
This fixes (hopefully) #121 and #107 by properly overriding AbstractAction::start method to allow updating current plan without stopping controller execution.

The problem was that we accepted a new goal but kept using the initial goal handle, and so the last goal received before stopping got LOST (all previous ones get PREEMPTED, so no problem). With this PR, we always use the last received goal handle. Also I tried to simplify the code by storing in a single map (concurrency slots map) goal handle, execution, and working thread.

### DIFF
--- a/mbf_abstract_nav/include/mbf_abstract_nav/abstract_action.h
+++ b/mbf_abstract_nav/include/mbf_abstract_nav/abstract_action.h
@@ -40,7 +40,6 @@
 #define MBF_ABSTRACT_NAV__ABSTRACT_ACTION_H_
 
 #include <actionlib/server/action_server.h>
-#include <boost/bimap/bimap.hpp>
 #include <mbf_abstract_nav/MoveBaseFlexConfig.h>
 #include "mbf_abstract_nav/robot_information.h"
 
@@ -51,12 +50,17 @@ template <typename Action, typename Execution>
 class AbstractAction
 {
  public:
-  typedef boost::bimaps::bimap<uint8_t, std::string> SlotGoalIdMap;
-
   typedef boost::shared_ptr<AbstractAction> Ptr;
 
   typedef typename actionlib::ActionServer<Action>::GoalHandle GoalHandle;
   typedef boost::function<void (GoalHandle &goal_handle, Execution &execution)> RunMethod;
+
+  typedef struct
+  {
+    typename Execution::Ptr execution;
+    boost::thread* thread_ptr;
+    GoalHandle goal_handle;
+  } ConcurrencySlot;
 
   AbstractAction(
       const std::string& name,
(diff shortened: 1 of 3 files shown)


mbf_abstract_nav/include/mbf_abstract_nav/abstract_action.h
```diff
@@ -51,106 +50,109 @@ template <typename Action, typename Execution>
...
     }
-    concurrency_slots_.insert(SlotGoalIdMap::value_type(goal_handle.getGoal()->concurrency_slot, goal_handle.getGoalID().id));
-    executions_.insert(std::pair<const std::string, const typename Execution::Ptr>(goal_handle.getGoalID().id, execution_ptr));
-    threads_ptrs_.insert(std::pair<const std::string, boost::thread*>(goal_handle.getGoalID().id,
-                         threads_.create_thread(boost::bind(&AbstractAction::runAndCleanUp, this, goal_handle, execution_ptr))));
+    // fill concurrency slot with the new goal handle, execution, and working thread
+    concurrency_slots_[slot].goal_handle = goal_handle;
+    concurrency_slots_[slot].execution = execution_ptr;
+    concurrency_slots_[slot].thread_ptr = threads_.create_thread(boost::bind(&AbstractAction::runAndCleanUp,
+                                                                                 this, goal_handle, execution_ptr));
```
spuetz: If we here write `concurrency_slots_[slot]` instead of `goal_handle` we could avoid the map access in the run method, since it uses the reference version of goal_handle. Thereby it uses the object which lives in the struct. See https://docs.ros.org/melodic/api/actionlib/html/server__goal__handle__imp_8h_source.html#l00340 
Or am I missing something?

mbf_abstract_nav/src/controller_action.cpp
```diff
@@ -51,59 +51,38 @@ ControllerAction::ControllerAction(
...
-      }else{
-        elem->second->cancel();
-        concurrency_slots_.left.erase(slot->first);
-      }
+      // Goal requests to run the same controller on the same concurrency slot:
+      // we update the goal handle and pass the new plan to the execution without stopping it
+      execution_ptr = slot_it->second.execution;
+      execution_ptr->setNewPlan(goal_handle.getGoal()->path.poses);
+      concurrency_slots_[slot].goal_handle = goal_handle;
+      return;
```
spuetz, line 73: Use else instead -> more readable

mbf_abstract_nav/src/controller_action.cpp
```diff
@@ -126,7 +105,7 @@ void ControllerAction::run(GoalHandle &goal_handle, AbstractControllerExecution
   {
     fillExePathResult(geometry_msgs::PoseStamped(), geometry_msgs::PoseStamped(),
                       mbf_msgs::ExePathResult::INVALID_PATH, "Controller started with an empty plan!", result);
-    goal_handle.setAborted(result, result.message);
+    concurrency_slots_[slot].goal_handle.setAborted(result, result.message);
```
spuetz, line 110: See previous comment: We could use the benefit of references here. Same for the following lines with the map access.

mbf_abstract_nav/include/mbf_abstract_nav/abstract_action.h
```diff
@@ -51,106 +50,109 @@ template <typename Action, typename Execution>
...
     {
-      typename std::map<const std::string, const typename Execution::Ptr>::const_iterator elem
-          = executions_.find(slot->second);
-      if(elem != executions_.end())
-      {
-        elem->second->cancel();
-      }
-      concurrency_slots_.left.erase(slot->first);
+      // if there is a plugin running on the same slot, cancel it
+      slot_it->second.execution->cancel(); // TODO make thread safe
```
spuetz, line 83: Still not thread safe, should we do that in another PR or commit? `cancel` could reach the `CleanUp` part of `runAndCleanUp` after assigning the new thread pointer in line 88. This will be the case, since the mutex blocks until we finished assigning. Then it deletes the new thread instead of the old one, or am I wrong?

mbf_abstract_nav/include/mbf_abstract_nav/abstract_action.h
```diff
@@ -51,106 +50,109 @@ template <typename Action, typename Execution>
...
 
-  void runAndCleanUp(GoalHandle goal_handle, typename Execution::Ptr execution_ptr){
+  virtual void runAndCleanUp(GoalHandle &goal_handle, typename Execution::Ptr execution_ptr){
+    uint8_t slot = goal_handle.getGoal()->concurrency_slot;
+
     if (execution_ptr->setup_fn_)
       execution_ptr->setup_fn_();
-    run_(goal_handle, *execution_ptr);
-    ROS_DEBUG_STREAM("Finished action run method, waiting for execution thread to finish.");
+    run_(concurrency_slots_[slot].goal_handle, *execution_ptr);
```
spuetz, line 108: Aa here you use the map version of the `goal_handle`. Shouldn't it then already work with the reference version and not the map version in the run method?

spuetz (changes requested):
Nice approach, some little changes :) And maybe we could use the benefit of references instead of accessing the map for the current `goal_handle`.

mbf_abstract_nav/include/mbf_abstract_nav/abstract_action.h
corot: Not possible, because the references will keep "pointing" to the original GoalHandle (very well explained [here](https://www.walletfox.com/course/pointerref.php)). Should be possible storing and passing a pointer instead of a reference, but then we need to change the _run_ and _runAndCleanUp_ prototypes everywhere

corot: what... maybe is not a bad idea after all... :thinking: 
do u think I should try?

mbf_abstract_nav/src/controller_action.cpp
corot: mmm...  yes, but would be more complex structure, as I want to prevent the execution of the _AbstractAction::start(goal_handle, execution_ptr)_ if any of the ifs is false (so either 2 elses with the call repeated, or a flag, both uglier than the return, imho)

mbf_abstract_nav/include/mbf_abstract_nav/abstract_action.h
corot: Probably yes,,,  not thread safe. But I would rather fix in a separated PR, as this is already quite complex

mbf_abstract_nav/include/mbf_abstract_nav/abstract_action.h
corot: See my comment above. Unless we switch to pointers everywhere, we must stick to the map version (uglier, I know :see_no_evil: )

spuetz: I know the difference between ptr and ref. But see the documentation of the goal handle. The object stays the same. Only the internal fields are changed using the assign operator

corot: mmm...  you are right...  but I tried the change you propose and it stopped working! So I assume the run method keeps accessing the old GoalHandle. I need to investigate what's really going on here

mbf_abstract_nav/src/controller_action.cpp
spuetz: Maybe you need to define a ref in the struct and hand over the ref via a constructor.

mbf_abstract_nav/src/controller_action.cpp
```diff
@@ -131,160 +133,164 @@ void ControllerAction::run(GoalHandle &goal_handle, AbstractControllerExecution
...
-      controller_active = false;
-      fillExePathResult(robot_pose, goal_pose,
-                        mbf_msgs::ExePathResult::TF_ERROR, "Could not get the robot pose!", result);
-      concurrency_slots_[slot].goal_handle.setAborted(result, result.message);
-      ROS_ERROR_STREAM_NAMED(name_, result.message << " Canceling the action call.");
-      break;
-    }
-
-    if (first_cycle)
     {
```
spuetz, line 136: Use a method `getGoalHandle` where the slot is locked inside, returning a copy of the goal handle, or use manual mtx.lock(), mtx.unlock() here. This avoids this code remapping and can safely be used everywhere.

spuetz (changes requested):

mbf_abstract_nav/include/mbf_abstract_nav/abstract_action.h
corot: I have tracked the goal id and I can confirm that the overloaded assignment operator you mentioned doesn't affect already created references. Then, we have only two options: let the PR as it is now, or change all to pointers

spuetz: Maybe you need to define a ref in the struct and hand over the ref via a constructor. Could you try that as the last choice.

spuetz:
Is covered by #137

PR #137, opened 2019-06-11
 Refactor by passing only concurrency slot to run methods

corot:
The concurrency slots map is now private and accesed by get/set methods
that lock the slots mutex.


spuetz (changes requested):
I would suggest to have a class ConcurrencySlot with getters and setters and then `typedef boost::function<void (ConcurrencySlot& slot)> RunMethod;`

PR #141, opened 2019-06-17
On controller execution, check that local costmap is current

corot:
I'm thinking,,, maybe overridding as in this PR is cleaner than [passing setup and cleanup functions](https://github.com/magazino/move_base_flex/blob/master/mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp#L111) to the abstract execution :thinking:

### DIFF
--- a/mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
+++ b/mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
@@ -236,11 +236,20 @@ namespace mbf_abstract_nav
      */
     virtual void run();
 
+    /**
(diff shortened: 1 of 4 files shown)


spuetz:
maybe also even more abstract, something like a safety break callback function. Could be loaded as plugin. Or is that too much overhead?

corot:
mmm...  pluggin --> over-killing.   What about _isDriveSafe_? or _safeToDrive_?

spuetz:
What about  `safetyCheck()` or `controllerSafetyCheck()`?

corot:
I changed to isSafeToDrive.  No problem neither with safetyCheck

PR #148, opened 2019-07-27
Fix #147: Reset oscillation timer after executing a recovery behavior

### LINKED ISSUE #147: Oscillation timeout doesn't reset after recovery succeeded
This happens when using **Move Base action**. 
When oscillation is detected and recovery behaviors are executed, MBF doesn't give enough time to the robot to move and get out of the oscillating state. 

This causes that all recovery behaviors are fired and when none are left the goal is aborted. 
https://github.com/magazino/move_base_flex/blob/62b46461fd2a04f10dd33f0f9f947ce33baea193/mbf_abstract_nav/src/move_base_action.cpp#L240

`last_oscillation_reset_` and `last_oscillation_pose_` could be reset here and solve the issue.
https://github.com/magazino/move_base_flex/blob/62b46461fd2a04f10dd33f0f9f947ce33baea193/mbf_abstract_nav/src/move_base_action.cpp#L474

corot:


mbf_abstract_nav/src/move_base_action.cpp
```diff
@@ -230,10 +230,7 @@ void MoveBaseAction::actionExePathFeedback(
       {
         mbf_msgs::MoveBaseResult move_base_result;
         move_base_result.outcome = mbf_msgs::MoveBaseResult::OSCILLATION;
-        if(recovery_enabled_)
```
spuetz, line 233: Why removing the `recovery_enabled` option?

corot: We don't mention recovery behaviors on other failures message (planner/controller failures). So why to do on OSCILLATION?

Issue #109, opened 2018-11-19
Publish zero velocity when reached to goal. 


spuetz:
Publish zero velocity when reached to goal. If the plugin is not publishing zero velocity to `cmd_vel` the robot will go one with the last command. This bug is in particular caused when setting the `mbf_tolerance_check` to true since the plugin will not publish zero velocity in most of the cases.  Related to #32 and #70.

howardcochran:
I request that this be made optional. Some controllers provide the ability to arrive at a goal at full speed as a feature.  For example, TebLocalPlanner, when its free_goal_vel parameter set to true. This makes it possible to immediately give it a new goal so that it can go from goal to goal without stopping -- rolling through waypoints.  If MBF hard-coded a zero twist, it would break this capability.

spuetz:
Japp you're right!

corot:
From #171:

Would it be okay if this becomes the default if controller_->cancel() fails?
If it succeeds it would be up to the user to publishZeroVelocity() if they desire so.

Or perhaps always leave it up to the user to implement this.

I would say the last option is the simplest and more predictable, defaulting to pub the zero vel. @spuetz , your thoughts?
For para name: force_stop_on_cancel = true sounds good?

corot:
This got implemented with #166, right?

PR #166, opened 2019-12-26
add parameter to actively stop once the goal is reached

ChristofDubs:
Related to Issue #109, the current implementation (correct me if I'm wrong) requires one to delegate the responsibility to stop the robot to the local planner. If someone wanted to ensure that the robot stopped at a goal, one would have to do the following:
- The local planner has to return false for `isGoalReached()` if the robot isn't stopped yet (i.e. if last call to `computeVelocityCmd()` did not set `cmd_vel` to zero)
- `mbf_tolerance_check_` needs to be false (otherwise above doesn't work)
- You can probably not use multiple local planners, because this method relies on the planner's last internally calculated velocity command being the last one that was published by move_base_flex

This is not very nice; as the thread in the issue suggests, this should be handled by a configurable parameter.

I verified on my system that adding these changes stops the robot at the goal (by modifying the local planner to return non-zero commands in all cases)

### DIFF
--- a/mbf_abstract_nav/src/abstract_controller_execution.cpp
+++ b/mbf_abstract_nav/src/abstract_controller_execution.cpp
@@ -65,7 +65,7 @@ namespace mbf_abstract_nav
     private_nh.param("robot_frame", robot_frame_, std::string("base_link"));
     private_nh.param("map_frame", global_frame_, std::string("map"));
     private_nh.param("mbf_tolerance_check", mbf_tolerance_check_, false);
-    private_nh.param("stop_at_goal", stop_at_goal_, true);
+    private_nh.param("force_stop_at_goal", force_stop_at_goal_, true);
     private_nh.param("dist_tolerance", dist_tolerance_, 0.1);
     private_nh.param("angle_tolerance", angle_tolerance_, M_PI / 18.0);
     private_nh.param("tf_timeout", tf_timeout_, 1.0);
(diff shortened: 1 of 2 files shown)


corot:
Makes sense. And code lgtm. Maybe rename "stop_at_goal" as "force_stop_at_goal"?  Cause "stop_at_goal" sounds like if false, the robot won't stop!
Also I would change default to false, so current behavior doesn't change unless actively setting the new param.

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -65,6 +65,7 @@ namespace mbf_abstract_nav
     private_nh.param("robot_frame", robot_frame_, std::string("base_link"));
     private_nh.param("map_frame", global_frame_, std::string("map"));
     private_nh.param("mbf_tolerance_check", mbf_tolerance_check_, false);
+    private_nh.param("stop_at_goal", stop_at_goal_, true);
```
spuetz: change default to `false`

spuetz (changes requested):

ChristofDubs:
> Also I would change default to false, so current behavior doesn't change unless actively setting the new param.

I get the point of not changing the current behavior. The reason I picked `true` as default was that I think this is more close to the default behavior a (new) user would expect, also the described advanced use-case to not publish a zero velocity command to "roll through waypoints" seems like a workaround to me for a via_points feature (which the mentioned TEB local planner actually already supports).

Anyway I changed the default to `false`, you can merge it like that, otherwise let me know and I'd be happy to revert back the change.

corot:
Thanks!

Issue #171, opened 2020-02-13
Graceful cancel (preempt)


Timple:
When preempting the "exe_path" action we would like to come to a graceful stop and then say we succeeded in cancelling.

### Current implementation:
 sets the `cancel_` flag op [true](https://github.com/magazino/move_base_flex/blob/master/mbf_abstract_nav/src/abstract_controller_execution.cpp#L236) before we can handle [any actions](https://github.com/magazino/move_base_flex/blob/master/mbf_abstract_nav/src/abstract_controller_execution.cpp#L238) in our overloaded cancel function.

This results in an immediate [zero publish](https://github.com/magazino/move_base_flex/blob/master/mbf_abstract_nav/src/abstract_controller_execution.cpp#L273) on the `cmd_vel` topic.

### Desired behavior:
Allow for a custom implementation upon receiving a preempt on an "exe_path" action. For example: We want to adhere to deceleration limits and still follow the current active path. So it could also take some time before the cancel is acknowledged.

corot:
Sounds reasonable; did you try to move the
`  cancel_ = true;
`
to lines 242 and 244 just brefore both returns? Then the cancel call to your plugin will activate your graceful stop mode, during which you provide slowing down velocities until you are ready to stop. Then you return the cancel call, and MBF will stop the robot and terminate the exe_path action. All this should be possible because the cancel runs in a separated thread from the main execution.

Timple:
I tried to move the line and that seemed to work. I was surprised they run in a separated thread.

However still a goal of 0.0 is send at the end. Which might or might not be what somebody wants.
We could also leave this up to the user (and maybe default it if no cancel implementation is available).

corot:
Can make [this line](https://github.com/magazino/move_base_flex/commit/e4cdbb9f9ca0191cfd1905ff5125f67685292534#diff-7bfab1323bf81df50a6c81b1a4965dacR272) parametrizable, yes. We added on request by another user to ensure that the robot doesn't keep moving at a fixed speed if the controller dies

Timple:
Would it be okay if this becomes the default if `controller_->cancel()` fails?
If it succeeds it would be up to the user to `publishZeroVelocity()` if they desire so.

Or perhaps always leave it up to the user to implement this.

corot:
> Would it be okay if this becomes the default if `controller_->cancel()` fails?
> If it succeeds it would be up to the user to `publishZeroVelocity()` if they desire so.
> 
[...]

PR #172, opened 2020-02-14
Allow the user time-consuming cancel implementations

Timple:
by setting the flag after function succeeds

Solves #171

### DIFF
--- a/mbf_abstract_nav/src/abstract_controller_execution.cpp
+++ b/mbf_abstract_nav/src/abstract_controller_execution.cpp
@@ -234,14 +234,14 @@ bool AbstractControllerExecution::reachedGoalCheck()
 bool AbstractControllerExecution::cancel()
 {
   // returns false if cancel is not implemented or rejected by the recovery behavior (will run until completion)
-  if(!controller_->cancel())
+  bool ctrl_cancelled = controller_->cancel();
+  if(!ctrl_cancelled)
   {
     ROS_WARN_STREAM("Cancel controlling failed or is not supported by the plugin. "
                         << "Wait until the current control cycle finished!");
-    return false;
   }
   cancel_ = true;
-  return true;
+  return ctrl_cancelled;
 }
 
 
--- a/mbf_abstract_nav/src/abstract_controller_execution.cpp
+++ b/mbf_abstract_nav/src/abstract_controller_execution.cpp
@@ -233,7 +233,6 @@ bool AbstractControllerExecution::reachedGoalCheck()
 
 bool AbstractControllerExecution::cancel()
 {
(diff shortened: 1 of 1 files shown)


mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -233,14 +233,14 @@ bool AbstractControllerExecution::reachedGoalCheck()
 
 bool AbstractControllerExecution::cancel()
 {
-  cancel_ = true;
   // returns false if cancel is not implemented or rejected by the recovery behavior (will run until completion)
   if(!controller_->cancel())
   {
     ROS_WARN_STREAM("Cancel controlling failed or is not supported by the plugin. "
                         << "Wait until the current control cycle finished!");
     return false;
```
corot: We need the `cancel_ = true;` also here. Can you get the result of cancel() and return it after setting cancel_ to true?

```
bool AbstractControllerExecution::cancel()
{
  // returns false if cancel is not implemented or rejected by the recovery behavior (will run until completion)
  bool ctrl_cancelled = controller_->cancel();
  if(!ctrl_cancelled)
  {
    ROS_WARN_STREAM("Cancel controlling failed or is not supported by the plugin. "
                        << "Wait until the current control cycle finished!");
  }
  cancel_ = true;
  return ctrl_cancelled;
}
```

corot (changes requested):

mbf_abstract_nav/src/abstract_controller_execution.cpp
Timple: Ah right, this was before my first coffee.
Something like this?
```
bool AbstractControllerExecution::cancel()
{
  // returns false if cancel is not implemented or rejected by the recovery behavior (will run until completion)
  bool can_cancel = controller_->cancel();
  ROS_WARN_STREAM_COND(!can_cancel, "Cancel controlling failed or is not supported by the plugin. "
                                        << "Wait until the current control cycle finished!");
  cancel_ = true;
  return can_cancel;
}
```

corot (approved):

corot:
lgtm, will test during the weekend

spuetz:
This is fine, logically it just moves the setting of the flag down, which will - as discussed - finish the cancel method first, and if the cancel method is not implemented, the controller execution will cancel the action after it returned from computeVel...

Timple:
Would this feature be enough to justify a new debian release? :innocent:

spuetz:
Sure, I will merge the other two open PR at the weekend and make a new release. ;) cheers.

Timple:
ping :)

PR #173, opened 2020-02-17
Add parameter force_stop_on_cancel to send a zero-speed command on canceling navigation

corot:
Following discussion on e4cdbb9f9ca0191cfd1905ff5125f67685292534 and #171.

Actually, I'm realizing that we have not documented all these parameters... So won't be usefull unless inspecting the code, what is probably a lot to ask.


spuetz:
Shouldn't the default of `force_stop_on_cancel` be `true` to keep the same behaviour as before?

corot:
> Shouldn't the default of `force_stop_on_cancel` be `true` to keep the same behaviour as before?

Well, we added that stop on 16 October: https://github.com/magazino/move_base_flex/commit/e4cdbb9f9ca0191cfd1905ff5125f67685292534#diff-7bfab1323bf81df50a6c81b1a4965dacR272

So the default behavior has been same as `false` for much longer. And doing nothing is typically a more predictable behavior

PR #179, opened 2020-04-01
use shared_from_this to avoid crash on shutdown

dorezyuk:
Hi guys, 

we had a crash on shutdown with following stack trace

> Stack trace of thread 3423:
> #0  0x00007f2f89d7ee97 __GI_raise (libc.so.6)
> #1  0x00007f2f89d80801 __GI_abort (libc.so.6)
> #2  0x00007f2f8a3d5957 n/a (libstdc++.so.6)
> #3  0x00007f2f8a3dbae6 n/a (libstdc++.so.6)
> #4  0x00007f2f8a3dab49 n/a (libstdc++.so.6)
> #5  0x00007f2f8a3db4b8 __gxx_personality_v0 (libstdc++.so.6)
> #6  0x00007f2f8a141573 n/a (libgcc_s.so.1)
> #7  0x00007f2f8a141df5 _Unwind_Resume (libgcc_s.so.1)
> #8  0x00007f2f8b12d4c9 _ZN5boost11unique_lockINS_5mutexEE4lockEv (libtf2_ros.so)
> #9  0x00007f2f8adf2e5b _ZN3ros12TimerManagerINS_4TimeENS_8DurationENS_10TimerEventEE6removeEi (libroscpp.so)
> #10 0x00007f2f8adf1c3b _ZN3ros5Timer4Impl4stopEv (libroscpp.so)
> #11 0x00007f2f8adf1cfb _ZN3ros5Timer4ImplD2Ev (libroscpp.so)
> #12 0x00007f2f8adf3e82 _ZN5boost6detail17sp_counted_impl_pIN3ros5Timer4ImplEE7disposeEv (libroscpp.so)
> #13 0x00007f2f8adf1bd1 _ZN3ros5TimerD2Ev (libroscpp.so)
> #14 0x00007f2f8b426999 _ZN15mbf_costmap_nav14CostmapWrapperD1Ev (libmbf_costmap_server.so)
> #15 0x00007f2f8b4269c9 _ZN15mbf_costmap_nav14CostmapWrapperD0Ev (libmbf_costmap_server.so)
> #16 0x00005560e651c28a _ZN5boost6detail15sp_counted_base7releaseEv (mbf_costmap_nav)
> #17 0x00007f2f8b3cbe47 _ZN15mbf_costmap_nav23CostmapNavigationServerD1Ev (libmbf_costmap_server.so)
[...]

PR #180, opened 2020-04-01
Ensure that check_costmap_mutex_ is destroyed after timer.

MichaelGrupp:
Reorders the member declaration so that the timer event can't attempt to
lock the destroyed mutex.

### DIFF
--- a/mbf_costmap_nav/include/mbf_costmap_nav/costmap_wrapper.h
+++ b/mbf_costmap_nav/include/mbf_costmap_nav/costmap_wrapper.h
@@ -107,12 +107,12 @@ private:
   //! Private node handle
   ros::NodeHandle private_nh_;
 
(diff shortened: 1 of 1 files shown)


corot (commented):
Interesting. When happens this? How can I reproduce?

So this work cause the attributes are piled in the stack as they are declared, and destroyed in the inverse order, right?

MichaelGrupp:
> Interesting. When happens this? How can I reproduce?

I noticed that while we were debugging the problem described in Dima's PR #179
This patch here doesn't fix the whole issue, but it's nevertheless good to pay attention to the resource lifetime here IMO.

> So this work cause the attributes are piled in the stack as they are declared, and destroyed in the inverse order, right?

Yes, that's how the C++ standard defines it. I already had a bug where this mattered ([here](https://github.com/googlecartographer/cartographer_ros/pull/1235)), so I'm a bit cautious with this in multi-threading contexts :thinking:

Issue #181, opened 2020-04-08
mbf_tolerance_check not dynamically reconfigurable


Slinky-Wrangle-Punch:
It would be nice if the mbf_tolerance check would be dynamically reconfigurable so that the tolerance check can be adjusted for different types of target poses.

The parameters of interest are mbf_tolerance_check, dist_tolerance and angle_tolerance:

https://github.com/magazino/move_base_flex/blob/c16d21b1a7364c960aa1bbae36221f4b7e77481d/mbf_abstract_nav/src/abstract_controller_execution.cpp#L69-L71

corot:
Sounds reasonable. But by other hand, all local planners I can think about have their own dynamically reconfigurable tolerance parameters, so those ones look redundant (and confusing, given the duplication). Moreover, those parameters cannot overrule controller's ones: look at this line: https://github.com/magazino/move_base_flex/blob/c16d21b1a7364c960aa1bbae36221f4b7e77481d/mbf_abstract_nav/src/abstract_controller_execution.cpp#L243 There's an `OR`. To make these parameters really useful, we shouldn't ask the controller if `mbf_tolerance_check` is true.

So I propose:
- remove these parameters
- add them to the action goal (same as planner)
- change the above line so this parameters overrule controller's ones
- modify planner action to have both dist and angle tolerances

Like that, we get a simpler mechanism to tailor controller behavior, w/o the need to call dyn. reconf. two times every time we want more or less precision than usual.
As a bonus, we get a nice parallelism between controller and planner.

Sounds reasonable? @spuetz, your thoughts?

spuetz:
What? The OR makes sense, there are two more lines ;) It means that either the controller plugin returns true, if the goal is reached with respect to the tolerance values or mbf says so, if activated. I suggested @mhoellmann to open this issue, since we have this issue with one of our robots, I will implement something solving this the next days. The `mbf_tolerance_check` param enables mbf to check for reached goal. This is necessary since many controllers have not implemented the tolerance check in `isGoalReached`. 
This parameter makes absolutely sense. I propose: 
- Adding the `angle_tolerance` and the `dist_tolerance` parameter to the action goal in `ExePath.action`.
- Overwriting the parameters in the corresponding controller execution -> implement this functionality in the `AbstractControllerExecution` class.

corot:
[...]

PR #182, opened 2020-04-16
mbf_tolerance check for missions

ebimor:
It would be valuable if we could set/change the goal tolerance for each mission that is submitted to the mbf. This is mainly to capture the knowledge of the human operator about the goal point environment. For instance, depending on the accuracy of the localization at the goal point area or whether it might be too many obstacles around/at goal point location, the operator can change the tolerance for a goal point. This change is backward compatible and turned off by default, i.e., if user does not specify a tolerance for the goal, mbf will use the default logic to check for goal_reached criteria

### DIFF
--- a/mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
+++ b/mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
@@ -111,7 +111,7 @@ namespace mbf_abstract_nav
      * @brief Sets a new plan to the controller execution
      * @param plan A vector of stamped poses.
      */
-    void setNewPlan(const std::vector<geometry_msgs::PoseStamped> &plan);
+    void setNewPlan(const std::vector<geometry_msgs::PoseStamped> &plan, bool has_tolerance = false, double dist_to_goal_tolerance = 0.0, double angle_to_goal_tolerance = 3.14);
 
     /**
      * @brief Cancel the planner execution. This calls the cancel method of the planner plugin. This could be useful if the
--- a/mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
+++ b/mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
@@ -358,13 +358,13 @@ namespace mbf_abstract_nav
     //! current robot pose;
     geometry_msgs::PoseStamped robot_pose_;
 
-    //! whether check for action specific has_tolerance
+    //! whether check for action specific tolerance
     bool has_tolerance_;
 
-    //! replaces default distance to goal tolerance for the action
+    //! replaces parameter dist_tolerance_ for the action
     double dist_to_goal_tolerance_;
 
-    //! replaces default angle to goal tolerance for the action
+    //! replaces parameter angle_tolerance_ for the action
     double angle_to_goal_tolerance_;
   };
 
--- a/mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
+++ b/mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
@@ -110,15 +110,15 @@ namespace mbf_abstract_nav
     /**
      * @brief Sets a new plan to the controller execution
      * @param plan A vector of stamped poses.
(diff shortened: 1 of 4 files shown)


spuetz:
Is related to #181

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -239,7 +246,14 @@ bool AbstractControllerExecution::isMoving()
 
 bool AbstractControllerExecution::reachedGoalCheck()
 {
-  // check whether the controller plugin returns goal reached or if mbf should check for goal reached.
+  //if action has a specific tolerance, check goal reached with those tolerances
+  if(has_tolerance_){
+    return controller_->isGoalReached(dist_to_goal_tolerance_, angle_to_goal_tolerance_) ||
```
spuetz: The `mbf_tolerance_check_` is missing here.

mbf_abstract_nav/src/controller_action.cpp
```diff
@@ -76,7 +76,10 @@ void ControllerAction::start(
       // Goal requests to run the same controller on the same concurrency slot:
       // we update the goal handle and pass the new plan to the execution without stopping it
```
spuetz: Add comments here.

mbf_msgs/action/ExePath.action
```diff
@@ -8,6 +8,10 @@ string controller
 # use different slots for concurrency
 uint8 concurrency_slot
 
+# define goal tolerance for the mission
+bool goal_tolerance
+float32 dist_to_goal_tolerance
+float32 angle_to_goal_tolerance
```
spuetz: Please use `dist_tolerance`, `angle_tolerance`.

mbf_msgs/action/ExePath.action
```diff
@@ -8,6 +8,10 @@ string controller
 # use different slots for concurrency
 uint8 concurrency_slot
 
+# define goal tolerance for the mission
+bool goal_tolerance
```
spuetz: Use `tolerance_from_action`, or something similar, otherwise this could be misunderstood.

mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
```diff
@@ -351,6 +358,14 @@ namespace mbf_abstract_nav
...
     geometry_msgs::PoseStamped robot_pose_;
 
+    //! whether check for action specific tolerance
+    bool has_tolerance_;
+
+    //! replaces parameter dist_tolerance_ for the action
+    double dist_to_goal_tolerance_;
+
+    //! replaces parameter angle_tolerance_ for the action
+    double angle_to_goal_tolerance_;
```
spuetz: Please use `action_dist_tolerance_`, `action_angle_tolerance_`, `tolerance_from_action_`.

spuetz (changes requested):
Thank you for this PR!

mbf_msgs/action/ExePath.action
ebimor: My idea was not to replace the original dist_tolerance and angle_tolerance in the mbf. So I defined these new variables to be specific for a mission. If a mission does not have tolerances, the default tolerances (i.e., dist_tolerance and angle_tolerance) will be used. What do you think?

ebimor: Maybe I can call them action_dist_tolerance and action_angle_tolerance?

corot:
A bit too late, but... I would have added the same fields on GetPath action; a nice symmetry.
Also, `tolerance-from-action` is a bit weird in the action itself... I would prefer `use-tolerances` (same way we have `use-start-pose` on GetPath action).
I can open a follow-up PR, if u want.

spuetz:
I have asked to change the variable to `tolerance_from_action` on purpose. Otherwise, it's ambiguous. But the symmetry argument, is a good one.

corot (commented):
> I have asked to change the variable to `tolerance_from_action` on purpose. Otherwise, it's ambiguous. But the symmetry argument, is a good one.

ok,,, but tolerance from goal is even clearer

Issue #183, opened 2020-04-21
extensive get_path requests make move_base_flex stop responding


ChristofDubs:
In rare cases, issuing a new `get_path` request immediately after receiving the result of the previous one results in a `Start action "get_path" using planner "GlobalPlanner" of type "global_planner/GlobalPlanner"` confirmation, immediately (a few milliseconds) followed by a `Cancel planning failed or is not supported by the plugin. Wait until the current planning finished!` warning, and rhen no more messages being published on `get_path/result`.

To reproduce this, a goal can be sent
```
void sendRequest() {
	ac->sendGoal(goal,
		[](const actionlib::SimpleClientGoalState& state, const mbf_msgs::GetPathResultConstPtr& result) {
			onResult(result);
		},
		[]() {}, [](const mbf_msgs::GetPathFeedbackConstPtr& feedback) {});
}
```
with a callback that re-sends the goal:
```
void onResult(const mbf_msgs::GetPathResultConstPtr& result) {
	if (!result) {
		ROS_WARN("Empty result.");
		return;
	}  
	sendRequest();
}
```

It took usually around 5 minutes of constant `get_path` requests until the move_base_flex became unresponsive; in parallel ran `sudo stress --cpu 4 --io 3 --vm 2 --vm-bytes 256M` to add some additional CPU load, but that may not be necessary to reproduce it.

spuetz:
All action goals on the same concurrency slot?

ChristofDubs:
> All action goals on the same concurrency slot?

Yes.

@corot already put together [this branch](https://github.com/magazino/move_base_flex/tree/get_path_freeze) that seems to fix it

corot:
fixed on master. @ChristofDubs, can you confirm, please?

ChristofDubs:
> fixed on master. @ChristofDubs, can you confirm, please?

The `get_path` behavior seems fine now. However, I am facing some more severe problem now:
When I send 2D nav goals (uses MoveBaseAction), it works fine as long as i send a new goal before the robot arrived. If I send a goal after the robot arrived, the robot doesn't move, and after starting the action, the following error appears (6ms later):
```
[ros.mbf_abstract_nav.exe_path]: Start action "exe_path" using controller "DWAPlannerROS" of type "dwa_local_planner/DWAPlannerROS"
[ros.mbf_abstract_nav.actionlib]: To transition to a cancelled state, the goal must be in a pending, recalling, active, or preempting state, it is currently in state: 3
```
[...]

PR #185, opened 2020-04-22
Issue #183: Reuse execution slots; cleanup only at destruction

corot:
Problem is that [here](https://github.com/magazino/move_base_flex/compare/get_path_freeze?expand=1#diff-21ac5e472653abd3e54d1178f443bb6eL88) we are joining a thread that can be already destroyed [here](https://github.com/magazino/move_base_flex/compare/get_path_freeze?expand=1#diff-21ac5e472653abd3e54d1178f443bb6eL83), because we do after unlocking slot_map_mtx_ [here](https://github.com/magazino/move_base_flex/compare/get_path_freeze?expand=1#diff-21ac5e472653abd3e54d1178f443bb6eL83).

With this PR, we don't cleanup execution slots, but reuse them (and cleanup remnants on shutdown).
I did heavy load with get_path/exe_path and the issue is fixed. I cannot detect neither any new memory or thread leak (though I noticed we have some at shutdown, se valgrind report below). I didn't perform stress testing with recovery actions.

```
==28807== HEAP SUMMARY:
==28807==     in use at exit: 688,798 bytes in 542 blocks
==28807==   total heap usage: 1,079,384 allocs, 1,078,842 frees, 570,985,052 bytes allocated
==28807== 
==28807== LEAK SUMMARY:
==28807==    definitely lost: 483,696 bytes in 33 blocks
==28807==    indirectly lost: 7,194 bytes in 122 blocks
==28807==      possibly lost: 304 bytes in 1 blocks
==28807==    still reachable: 197,604 bytes in 386 blocks
==28807==         suppressed: 0 bytes in 0 blocks
==28807== Rerun with --leak-check=full to see details of leaked memory
```

### DIFF
--- a/mbf_abstract_nav/include/mbf_abstract_nav/abstract_action_base.hpp
+++ b/mbf_abstract_nav/include/mbf_abstract_nav/abstract_action_base.hpp
@@ -54,8 +54,9 @@ class AbstractActionBase
   typedef boost::function<void (GoalHandle &goal_handle, Execution &execution)> RunMethod;
   typedef struct{
     typename Execution::Ptr execution;
-    boost::thread* thread_ptr;
+    boost::thread* thread_ptr = NULL;
     GoalHandle goal_handle;
+    bool in_use = false;
   } ConcurrencySlot;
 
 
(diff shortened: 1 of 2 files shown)


spuetz:
Maybe the memory leaks are in the costmaps or in the plugins, since we only use unique or shared pointers and the raw pointers we use are deleted, right?

corot:
> Maybe the memory leaks are in the costmaps or in the plugins, since we only use unique or shared pointers and the raw pointers we use are deleted, right?

Can be, yes

PR #189, opened 2020-04-24
Add getRobotVelocity function and mbf_msgs/ExePathFeedback/actual_vel

q576333:
I have Implemented getRobotVelocity function in the RobotInformation class and updateActualRobotVelocity function in AbstractControllerExecution class and tested. Let CostmapController plugin interface can get actual robot_velocity in computeVelocityCommands() function.
Also I add new ExePath action feedback variable "feedback_cmd_vel". I think maybe useful for some action client scenario. But I notice this feedback_cmd_vel have timing issue with the computeVelocityCommands() function's robot_velocity. 
In addition, There is some warning message when I ctrl + c the move_base_flex terminal. I need everyone help me to check my PR is correct or not?
The warning message is "Warning: class_loader.ClassLoader: SEVERE WARNING!!! Attempting to unload library while objects created by this loader exist in the heap! You should delete your objects before attempting to unload the library or destroying the ClassLoader. The library will NOT be unloaded. at line 108 in /tmp/binarydeb/ros-kinetic-class-loader-0.3.9/src/class_loader.cpp"

### DIFF
--- a/mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
+++ b/mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
@@ -49,6 +49,7 @@
 #include <tf/transform_listener.h>
 #include <geometry_msgs/PoseStamped.h>
 #include <geometry_msgs/Twist.h>
+#include <nav_msgs/Odometry.h>
 
 #include <mbf_utility/navigation_utility.h>
 #include <mbf_abstract_core/abstract_controller.h>
@@ -275,6 +276,11 @@ namespace mbf_abstract_nav
     bool computeRobotPose();
 
     /**
+     * @brief Update the robot actual velocity;
+     */
+    inline void updateActualRobotVelocity();
+
+    /**
      * @brief Sets the controller state. This method makes the communication of the state thread safe.
      * @param state The current controller state.
      */
@@ -357,6 +363,9 @@ namespace mbf_abstract_nav
 
     //! current robot pose;
     geometry_msgs::PoseStamped robot_pose_;
+
+    //! current robot actual velocity;
+    geometry_msgs::TwistStamped robot_velocity_;
 
     //! whether check for action specific tolerance
     bool tolerance_from_action_;
--- a/mbf_abstract_nav/include/mbf_abstract_nav/abstract_navigation_server.h
+++ b/mbf_abstract_nav/include/mbf_abstract_nav/abstract_navigation_server.h
@@ -347,6 +347,9 @@ typedef boost::shared_ptr<dynamic_reconfigure::Server<mbf_abstract_nav::MoveBase
     //! current_goal publisher for all controller execution objects
     ros::Publisher goal_pub_;
 
+    //! subscribe for the actual velocity feedback from odometry(wheel encoder)
+    ros::Subscriber odom_sub_;
+
     RobotInformation robot_info_;
 
     ControllerAction controller_action_;
--- a/mbf_abstract_nav/include/mbf_abstract_nav/controller_action.h
+++ b/mbf_abstract_nav/include/mbf_abstract_nav/controller_action.h
@@ -91,6 +91,7 @@ protected:
 
   boost::mutex goal_mtx_; ///< lock goal handle for updating it while running
   geometry_msgs::PoseStamped robot_pose_; ///< Current robot pose
+  geometry_msgs::TwistStamped robot_actual_vel_; //Actual robot velocity feedback
   geometry_msgs::PoseStamped goal_pose_;  ///< Current goal pose
 
 };
--- a/mbf_abstract_nav/src/abstract_navigation_server.cpp
+++ b/mbf_abstract_nav/src/abstract_navigation_server.cpp
@@ -72,6 +72,9 @@ AbstractNavigationServer::AbstractNavigationServer(const TFPtr &tf_listener_ptr)
   // init cmd_vel publisher for the robot velocity
   vel_pub_ = nh.advertise<geometry_msgs::Twist>("cmd_vel", 1);
 
+  // init odom_sub_ subscribe for the actual robot velocity
+  odom_sub_ = nh.subscribe<nav_msgs::Odometry>("odom", 1, &mbf_utility::actualVelCallback);
+
   action_server_get_path_ptr_ = ActionServerGetPathPtr(
     new ActionServerGetPath(
       private_nh_,
--- a/mbf_msgs/action/ExePath.action
+++ b/mbf_msgs/action/ExePath.action
@@ -57,3 +57,4 @@ float32 dist_to_goal
 float32 angle_to_goal
 geometry_msgs/PoseStamped  current_pose
 geometry_msgs/TwistStamped last_cmd_vel  # last command calculated by the controller
+geometry_msgs/TwistStamped actual_vel  # robot actual velocity feedback from wheel encoder
--- a/mbf_utility/include/mbf_utility/navigation_utility.h
+++ b/mbf_utility/include/mbf_utility/navigation_utility.h
@@ -42,6 +42,8 @@
 #define MBF_UTILITY__NAVIGATION_UTILITY_H_
 
 #include <geometry_msgs/PoseStamped.h>
+#include <geometry_msgs/TwistStamped.h>
+#include <nav_msgs/Odometry.h>
 #include <ros/duration.h>
 #include <ros/time.h>
 #include <string>
@@ -107,6 +109,20 @@ bool getRobotPose(const TF &tf,
                   const std::string &global_frame,
                   const ros::Duration &timeout,
                   geometry_msgs::PoseStamped &robot_pose);
+
+/**
+ * @brief subscribe odom_topic callback
+ * @param sub_odometry subscribe data from wheel odometry
+ */           
+void actualVelCallback(const nav_msgs::Odometry::ConstPtr& sub_odometry);
+
+/**
+ * @brief update the robot actual velocity.
+ * @param tf_listener TransformListener.
+ * @return true, if succeeded, false otherwise.
+ */
+void getRobotActualVelocity(geometry_msgs::TwistStamped &robot_velocity);
+
 /**
  * @brief Computes the Euclidean-distance between two poses.
  * @param pose1 pose 1
--- a/mbf_utility/src/navigation_utility.cpp
+++ b/mbf_utility/src/navigation_utility.cpp
@@ -42,6 +42,8 @@
 
 namespace mbf_utility
 {
+  geometry_msgs::TwistStamped actual_vel;
+  boost::mutex vel_mutex;
 
 bool getRobotPose(const TF &tf,
                   const std::string &robot_frame,
@@ -167,6 +169,19 @@ bool transformPoint(const TF &tf,
     return false;
   }
   return true;
+}
+
+void actualVelCallback(const nav_msgs::Odometry::ConstPtr& sub_odometry)
+{
+  boost::lock_guard<boost::mutex> lock(vel_mutex);
+  actual_vel.header = sub_odometry->header;
+  actual_vel.twist = sub_odometry->twist.twist;
+}
+
+void getRobotActualVelocity(geometry_msgs::TwistStamped &robot_velocity)
+{
+  boost::lock_guard<boost::mutex> lock(vel_mutex);
+  robot_velocity = actual_vel;
 }
 
 double distance(const geometry_msgs::PoseStamped &pose1, const geometry_msgs::PoseStamped &pose2)
--- a/mbf_abstract_nav/include/mbf_abstract_nav/robot_information.h
+++ b/mbf_abstract_nav/include/mbf_abstract_nav/robot_information.h
@@ -69,7 +69,7 @@ class RobotInformation{
    */
   bool getRobotPose(geometry_msgs::PoseStamped &robot_pose) const;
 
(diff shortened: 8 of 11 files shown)


mbf_abstract_nav/src/abstract_navigation_server.cpp
```diff
@@ -73,11 +73,16 @@ AbstractNavigationServer::AbstractNavigationServer(const TFPtr &tf_listener_ptr)
   oscillation_timeout_ = ros::Duration(oscillation_timeout);
   private_nh_.param("oscillation_distance", oscillation_distance_, 0.02);
 
+  private_nh_.param("odom_topic", odom_topic_, std::string("odom"));
```
corot: No param needed. Use the standard `odom` and can remap if different.

mbf_abstract_nav/include/mbf_abstract_nav/abstract_navigation_server.h
```diff
@@ -372,6 +375,9 @@ typedef boost::shared_ptr<dynamic_reconfigure::Server<mbf_abstract_nav::MoveBase
     //! current_goal publisher for all controller execution objects
     ros::Publisher goal_pub_;
 
+    //! subscribe for the actual velocity feedback from odometry(wheel encoder)
+    ros::Subscriber acutal_vel_sub_;
```
corot: why not just `odom_sub_`?

mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
```diff
@@ -358,6 +364,9 @@ namespace mbf_abstract_nav
     //! current robot pose;
     geometry_msgs::PoseStamped robot_pose_;
 
+    //! current robot actual velocity;
+    geometry_msgs::TwistStamped robot_velocity;
```
corot: Please respect convention: `robot_velocity_`

mbf_abstract_nav/include/mbf_abstract_nav/controller_action.h
```diff
@@ -91,6 +91,7 @@ class ControllerAction :
 
   boost::mutex goal_mtx_; ///< lock goal handle for updating it while running
   geometry_msgs::PoseStamped robot_pose_; ///< Current robot pose
+  geometry_msgs::TwistStamped robot_feedback_vel_; //Actual robot velocity feedback
```
corot: better `robot_vel_; ///< Actual robot velocity feedback`
(why feedback?;  and use inline docstring)

mbf_msgs/action/ExePath.action
```diff
@@ -57,3 +57,4 @@ float32 dist_to_goal
 float32 angle_to_goal
 geometry_msgs/PoseStamped  current_pose
 geometry_msgs/TwistStamped last_cmd_vel  # last command calculated by the controller
+geometry_msgs/TwistStamped feedback_cmd_vel  # robot actual velocity feedback from wheel encoder
```
corot: I don't like `feedback`... what about `current`? or `actual`?

mbf_utility/include/mbf_utility/navigation_utility.h
```diff
@@ -107,6 +109,23 @@ bool getRobotPose(const TF &tf,
...
                   const ros::Duration &timeout,
                   geometry_msgs::PoseStamped &robot_pose);
+/////////////////////////
+
+
+/**
+ * @brief subscribe odom_topic callback
+ * @param sub_odometry subscribe data from wheel odometry
+ */           
+void actual_vel_callback(const nav_msgs::Odometry::ConstPtr& sub_odometry);
```
corot: Please respect coding sttile: `actualVelCallback`

mbf_utility/src/navigation_utility.cpp
```diff
@@ -39,9 +39,13 @@
  */
 
 #include "mbf_utility/navigation_utility.h"
+//#include <iostream>
```
corot: ?

corot (changes requested):

mbf_msgs/action/ExePath.action
spuetz: What about `odom_vel_`?

corot:
Hi @q576333, are you still working on this change? Please rebase master and address the review comments if so.

As for the `Warning: class_loader.ClassLoader`, don't worry, it always happen when using plugins in ROS, not only on MBF.

q576333:
Sorry, recently I working hard my master thesis and also this is my first pull requests, so let me to understand how to rebase master, sorry again for keeping you waiting so long.

mbf_abstract_nav/src/abstract_navigation_server.cpp
q576333: Because I saw so many open source ros package also provide topic name parameter to let user change, but I think I can follow this package rule.

mbf_abstract_nav/include/mbf_abstract_nav/abstract_navigation_server.h
q576333: It's good recommend.

mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
q576333: I will change it.

mbf_abstract_nav/include/mbf_abstract_nav/controller_action.h
q576333: In control field normally call sensor information be a feedback. I am worried that will not be able to clearly indicate the meaning of this robot_vel_ in the program (like cmd_vel or actual velocity). maybe follow the same naming rule call robot_actual_vel_?

mbf_msgs/action/ExePath.action
q576333: I prefer odom_vel_ too. but I think ExePath action's feedback is for user to monitor each mobile robot behavior, so maybe follow whole naming rule, call actual_vel_ ?

mbf_utility/include/mbf_utility/navigation_utility.h
q576333: I will change it.

mbf_utility/src/navigation_utility.cpp
q576333: Original I want use c++ standard library's mutex, but I found this package already use the boost library, so I forgot to delete it.

mbf_abstract_nav/src/abstract_navigation_server.cpp
Rayman: Many packages are doing that, but I think that is an anti-pattern

mbf_msgs/action/ExePath.action
corot: `current_vel`, and move above last_cmd_vel;
mostly to match `current_pose`; both are physical properties:
```
geometry_msgs/PoseStamped  current_pose
geometry_msgs/TwistStamped current_vel   # robot actual velocity from the odometry
geometry_msgs/TwistStamped last_cmd_vel  # last command calculated by the controller
```
btw, given this interface change, and according to #191, this change won't be ever released :disappointed:

mbf_abstract_nav/include/mbf_abstract_nav/controller_action.h
corot: I would use `current_pose_` and `current_vel_`, as in the ExePath feedback msg, but put both attributes in the RobotInformation class (and move the odom topic there, too, as I explain in the review)

mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
```diff
@@ -274,6 +275,11 @@ namespace mbf_abstract_nav
      */
     bool computeRobotPose();
 
+    /**
+     * @brief Update the robot actual velocity;
+     */
+    inline void updateActualRobotVelocity();
```
corot, line 281: why inline?

mbf_utility/include/mbf_utility/navigation_utility.h
```diff
@@ -107,6 +109,20 @@ bool getRobotPose(const TF &tf,
                   const std::string &global_frame,
                   const ros::Duration &timeout,
                   geometry_msgs::PoseStamped &robot_pose);
+
+/**
+ * @brief subscribe odom_topic callback
+ * @param sub_odometry subscribe data from wheel odometry
+ */           
+void actualVelCallback(const nav_msgs::Odometry::ConstPtr& sub_odometry);
```
corot, line 117: The callback is for the odom topic, so... should be `odometryCallback`

mbf_abstract_nav/src/abstract_navigation_server.cpp
```diff
@@ -72,6 +72,9 @@ AbstractNavigationServer::AbstractNavigationServer(const TFPtr &tf_listener_ptr)
   // init cmd_vel publisher for the robot velocity
   vel_pub_ = nh.advertise<geometry_msgs::Twist>("cmd_vel", 1);
 
+  // init odom_sub_ subscribe for the actual robot velocity
+  odom_sub_ = nh.subscribe<nav_msgs::Odometry>("odom", 1, &mbf_utility::actualVelCallback);
```
corot, line 76: Why not subscribing on mbf_utility/include/mbf_utility/navigation_utility.h, if the callback is there?

corot (changes requested):
Still need some work. 
Moreover, seems clear to me that the odom subscription should be in RobotInformation class. And that it should be updated on demand, so the updateActualRobotVelocity new method is not necessary. But we can let this for a further refactoring.

mbf_abstract_nav/include/mbf_abstract_nav/controller_action.h
corot: btw, according to #191, this change won't be ever released :disappointed:

mbf_utility/src/navigation_utility.cpp
```diff
@@ -42,6 +42,8 @@
 
 namespace mbf_utility
 {
+  geometry_msgs::TwistStamped actual_vel;
+  boost::mutex vel_mutex;
```
Bitfroest, line 46: Just curious why no [std::mutex](https://en.cppreference.com/w/cpp/thread/mutex)?
Maybe name it `<full_name>_mutex`

Bitfroest: include is missing

mbf_utility/include/mbf_utility/navigation_utility.h
```diff
@@ -107,6 +109,20 @@ bool getRobotPose(const TF &tf,
...
+
+/**
+ * @brief update the robot actual velocity.
+ * @param tf_listener TransformListener.
+ * @return true, if succeeded, false otherwise.
```
FabianSchurig, line 122: outdated

mbf_utility/include/mbf_utility/navigation_utility.h
```diff
@@ -107,6 +109,20 @@ bool getRobotPose(const TF &tf,
...
+ */
+void getRobotActualVelocity(geometry_msgs::TwistStamped &robot_velocity);
```
Bitfroest, line 124: Just as note. A nice way is using `std::optional` in case of C++17 or boost::optional.

Do you still stick with 
```
void getRobotActualVelocity(geometry_msgs::TwistStamped &robot_velocity);

// instead of
std::optional<geometry_msgs::TwistStamped> getOdometryTwistStamped();
```

mbf_msgs/action/ExePath.action
spuetz: Jap, this line should be removed here. So we can not change the messages anymore. For noetic I think this will be ok, if we hurry a bit, but not for melodic and before.

mbf_msgs/action/ExePath.action
```diff
@@ -57,3 +57,4 @@ float32 dist_to_goal
 float32 angle_to_goal
 geometry_msgs/PoseStamped  current_pose
 geometry_msgs/TwistStamped last_cmd_vel  # last command calculated by the controller
+geometry_msgs/TwistStamped actual_vel  # robot actual velocity feedback from wheel encoder
```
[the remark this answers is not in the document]
spuetz: Jap, this line should be removed here. So we can not change the messages anymore. For noetic I think this will be ok, if we hurry a bit, but not for melodic and before. See #191

spuetz (changes requested):
Please remove the change from the action message for now. I would also prefer to not use the Robot Information. Keep things simple here: Add the subscriber to the `AbstractControllerExecution`. This also has a hidden advantage of the namespace and using different odom topics.

corot:
> Please remove the change from the action message for now. I would also prefer to not use the Robot Information. Keep things simple here: Add the subscriber to the `AbstractControllerExecution`. This also has a hidden advantage of the namespace and using different odom topics.

Why not concentrating all robot's state-related code on RobotInformation (and keep the executions simple)? We already handled current pose there, so... handling velocity too sounds quite reasonable to me. But maybe we are making poor @q576333 dizzy... we can merge this with just the requested changes and refactor later with whatever we decide.

spuetz:
But we are just receiving an odom message and hand it over to the plugins, I think it is a bit overkill and it looks hacky from my point of view. Furthermore, considering one mbf instance for multiple robots in the future a separate namespace is beneficial.

spuetz:
I'll implement this in the next week and open a new PR, then we can decide which one is better.

spuetz:
I'll address this in another PR.

PR #194, opened 2020-05-19
Avoid duplicated warn log msgs when we cannot cancel a plugin

corot:
```
[/move_base_flex  WARN 1589876693.431091917, 40.500000000]: Cancel recovery behavior "clearing_rotation" failed or is not supported by the plugin. Wait until the current recovery behavior finished!
[/move_base_flex  WARN 1589876693.431207950, 40.500000000]: Cancel recovering "clearing_rotation" failed or not supported; maybe wait until it is finished!
```
I dropped the 2nd one.

### DIFF
--- a/mbf_abstract_nav/src/recovery_action.cpp
+++ b/mbf_abstract_nav/src/recovery_action.cpp
@@ -83,10 +83,7 @@ void RecoveryAction::run(GoalHandle &goal_handle, AbstractRecoveryExecution &exe
         if (execution.isPatienceExceeded())
         {
           ROS_INFO_STREAM("Recovery behavior \"" << goal.behavior << "\" patience exceeded! Cancel recovering...");
-          if (!execution.cancel())
-          {
-            ROS_WARN_STREAM("Cancel recovering \"" << goal.behavior << "\" failed or not supported; maybe wait until it is finished!");
-          }
+          execution.cancel();
         }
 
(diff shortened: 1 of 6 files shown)


mbf_abstract_nav/src/recovery_action.cpp
```diff
@@ -83,7 +83,7 @@ void RecoveryAction::run(GoalHandle &goal_handle, AbstractRecoveryExecution &exe
         if (execution.isPatienceExceeded())
         {
           ROS_INFO_STREAM("Recovery behavior \"" << goal.behavior << "\" patience exceeded! Cancel recovering...");
-          if (!execution.cancel())
+          execution.cancel();
           {
             ROS_WARN_STREAM("Cancel recovering \"" << goal.behavior << "\" failed or not supported; maybe wait until it is finished!");
           }
```
spuetz: These lines should also be removed, right?

spuetz (changes requested):
A "cancel" could also be supported but not successful. So in such a case, there is no message at all with these changes, right?

mbf_abstract_nav/src/recovery_action.cpp
corot: :see_no_evil:

corot:
> A "cancel" could also be supported but not successful. So in such a case, there is no message at all with these changes, right?

No, it returns false and so we show the same msg: https://github.com/magazino/move_base_flex/pull/194/files#diff-2785ff456a3fe55e2adf26378b5353ddR198

PR #197, opened 2020-06-11
Hotfix: Prevent unrelated type casts for Cell

FabianSchurig:
Hello @corot and @spuetz,

An `'unsigned int'` value was c-style casted to an unrelated type `'uint16_t'` (aka `'unsigned short'`). The width of both data models might differ on platforms. Thus, memory is accessed out-of-bounds which results in non-deterministic return values of the `CheckPath` service and probably other methods of the `FootprintHelper`. 

I created two small examples demonstrating the issue with godbolt:

https://godbolt.org/z/rxpvjM
https://godbolt.org/z/X-pibJ

Best Fabian


corot (approved):
LGTM; thanks!

PR #198, opened 2020-06-15
Unify style

corot:
- Open brackets in its own line
- Make reference symbol position consistent across the project

### DIFF
--- a/mbf_abstract_nav/include/mbf_abstract_nav/abstract_action_base.hpp
+++ b/mbf_abstract_nav/include/mbf_abstract_nav/abstract_action_base.hpp
@@ -43,7 +43,8 @@
 #include <mbf_abstract_nav/MoveBaseFlexConfig.h>
 #include <mbf_utility/robot_information.h>
 
-namespace mbf_abstract_nav{
+namespace mbf_abstract_nav
+{
 
 template <typename Action, typename Execution>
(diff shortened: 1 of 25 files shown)


FabianSchurig:
Which style guide does mbf follow and how do you auto format? 

If there is no clang-format you could use https://github.com/davetcoleman/roscpp_code_format as suggested ín http://wiki.ros.org/CppStyleGuide . Unless you already use another style guide?

spuetz:
@corot  Can you resolve the conflicts and give an answer to the style guide question? I would say, we are more or less following the ROS style guide in a not strict way.

corot:
> Which style guide does mbf follow and how do you auto format?
> 
> If there is no clang-format you could use https://github.com/davetcoleman/roscpp_code_format as suggested ín http://wiki.ros.org/CppStyleGuide . Unless you already use another style guide?

Hi, I just accepted the suggestions from CLion; and for the reference symbol position, I just tried to make consistent across the project. As @spuetz said, we try to more-or-less follow ROS conventions, but we didn't a set a clang format to enforce it. Me in particular, I always use CLion, who formats the code as you type. I prefer that over automatic formatting, that often destroys sensible code alignments, nice-looking and easy-to-read for humans.

But I'm good with formally adopting the ROS format you mention (I can use it on CLion).

PR #199, opened 2020-06-18
fix deallocation on shutdown

dorezyuk:
Hey guys, we had issues with bad noisy shutdowns.

Here is what happens: The CostmapNavigaitonServer owns the classLoader for every plugin. Its base-class holds the the loaded plugins. When the destructor of the CostmapNavigaitonServer is called, the classLoaders go out of scope before the loaded plugins. 

The solution is to destruct explicitly all loaded plugins of inside the CostmapNavigaitonServer destructor - allowing a clear shutdown.

### DIFF
--- a/mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
+++ b/mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
@@ -89,6 +89,30 @@ CostmapNavigationServer::CostmapNavigationServer(const TFPtr &tf_listener_ptr) :
 
 CostmapNavigationServer::~CostmapNavigationServer()
 {
+  // remove every plugin before its classLoader goes out of scope.
+  {
+    const auto& names = planner_plugin_manager_.getLoadedNames();
+    for(const auto& name : names) {
+      planner_plugin_manager_.getPlugin(name).reset();
+    }
+  }
+  {
+    const auto& names = controller_plugin_manager_.getLoadedNames();
+    for(const auto& name : names) {
+      controller_plugin_manager_.getPlugin(name).reset();
+    }
+  }
+  {
+    const auto& names = recovery_plugin_manager_.getLoadedNames();
+    for(const auto& name : names) {
+      recovery_plugin_manager_.getPlugin(name).reset();
+    }
+  }
+
+  action_server_recovery_ptr_.reset();
+  action_server_exe_path_ptr_.reset();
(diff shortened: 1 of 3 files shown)


mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
```diff
@@ -89,6 +89,30 @@ CostmapNavigationServer::CostmapNavigationServer(const TFPtr &tf_listener_ptr) :
 
 CostmapNavigationServer::~CostmapNavigationServer()
 {
+  // remove every plugin before its classLoader goes out of scope.
+  {
+    const auto& names = planner_plugin_manager_.getLoadedNames();
```
spuetz: Please use the full type here for older C++ versions.

mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
```diff
@@ -89,6 +89,30 @@ CostmapNavigationServer::CostmapNavigationServer(const TFPtr &tf_listener_ptr) :
...
+    for(const auto& name : names) {
+      planner_plugin_manager_.getPlugin(name).reset();
+    }
+  }
+  {
+    const auto& names = controller_plugin_manager_.getLoadedNames();
```
spuetz: Please use the full type here for older C++ versions.

mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
```diff
@@ -89,6 +89,30 @@ CostmapNavigationServer::CostmapNavigationServer(const TFPtr &tf_listener_ptr) :
...
+    for(const auto& name : names) {
+      controller_plugin_manager_.getPlugin(name).reset();
+    }
+  }
+  {
+    const auto& names = recovery_plugin_manager_.getLoadedNames();
```
spuetz: Please use the full type here for older C++ versions.

spuetz (changes requested):

dorezyuk:
sorry the old approach didn't work - not sure how it worked this morning, since in the old code, I would only reset the new copy of the plugin and not the internally stored at the plugin manager.

Anyhow, I didn't found a way to solve the issue while maintaining ABI compatibility, and added a new clearPlugins method to get rid of the stuff stored in the manager

PR #204, opened 2020-07-03
Make robot_info a reference to keep one single instance

corot:
Also add attribute comment


spuetz:
Can you explain why this is necessary?

corot:
> Can you explain why this is necessary?

Sorry, prevents an unnecessary copy of `mbf_utility::RobotInformation`

PR #206, opened 2020-07-04
Do not call isGoalReached on controller when mbf_tolerance_check parameter is true

corot:
With this PR, we check if we have arrived instead of asking the controller. Current code considers that we have arrived when EITHER the controller or MBF consider that we have arrived, what is much less useful and confusing. You cannot, for example, enforce a higher precision than usual without calling controller dynamic reconfigure (assuming it has).

Caveat: as described in #213, MBF won't check if the robot is stopped when `check_goal_reached` is true, what is dangerous. As explained in the issue, it would be better instead to always delegate the arrived check to the controller, but pass him the tolerances provided in the goal when requested.

I also rename `mbf_tolerance_check` as `check_goal_reached`, as we are checking if we have arrived, independently of tolerances.

### DIFF
--- a/mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
+++ b/mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
@@ -311,13 +311,13 @@ namespace mbf_abstract_nav
     //! the last set plan which is currently processed by the controller
     std::vector<geometry_msgs::PoseStamped> plan_;
 
-    //! the duration which corresponds with the controller frequency.
+    //! the duration which corresponds with the controller frequency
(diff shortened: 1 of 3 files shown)


spuetz:
Hmmm this is now used in several projects... I would prefer to keep the name.

corot:
> Hmmm this is now used in several projects... I would prefer to keep the name.

Keep issue #213 in mind... if using this parameter, you can finish way off of the goal, as we are not checking if the robot is stopped at the goal. I'm mostly in favor of removing this feature entirely.

corot:
> Hmmm this is now used in several projects... I would prefer to keep the name.

Yup, that's why we do for noetic

spuetz (changes requested):

PR #218, opened 2020-08-14
Send move_base result when canceled during exe_path and recovery

corot:
On cancelling move_base action, we were calling goal_handle_.setCanceled only while running get_path, but not during exe_path nor recovery. So the action client didn't get any result (I noticed that because SMACH simple action state freezes after sending a cancel)

Other related changes (sorry, was too cumbersome to make a separated PR):
* Cancel whatever is running on oscillation (we can be also planning or recovering)
* Refactored a bit the code

Easily tested by sending any goal with axclient and canceling it, cause the status bar gets frozen in pink with the msg "Canceling goal":

`rosrun actionlib axclient.py /move_base_flex/move_base`

### DIFF
--- a/mbf_abstract_nav/src/move_base_action.cpp
+++ b/mbf_abstract_nav/src/move_base_action.cpp
@@ -102,18 +102,22 @@ void MoveBaseAction::reconfigure(
 void MoveBaseAction::cancel()
 {
   action_state_ = CANCELED;
+  cancelActiveClients();
+}
 
-  if(!action_client_get_path_.getState().isDone())
+void MoveBaseAction::cancelActiveClients()
+{
+  if (!action_client_get_path_.getState().isDone())
   {
     action_client_get_path_.cancelGoal();
   }
 
-  if(!action_client_exe_path_.getState().isDone())
+  if (!action_client_exe_path_.getState().isDone())
   {
     action_client_exe_path_.cancelGoal();
   }
 
-  if(!action_client_recovery_.getState().isDone())
+  if (!action_client_recovery_.getState().isDone())
   {
     action_client_recovery_.cancelGoal();
   }
@@ -148,9 +152,8 @@ void MoveBaseAction::start(GoalHandle &goal_handle)
   recovery_behaviors_ = goal.recovery_behaviors.empty() ? behaviors_ : goal.recovery_behaviors;
   current_recovery_behavior_ = recovery_behaviors_.begin();
 
-  geometry_msgs::PoseStamped robot_pose;
   // get the current robot pose only at the beginning, as exe_path will keep updating it as we move
-  if (!robot_info_.getRobotPose(robot_pose))
+  if (!robot_info_.getRobotPose(robot_pose_))
   {
     ROS_ERROR_STREAM_NAMED("move_base", "Could not get the current robot pose!");
     move_base_result.message = "Could not get the current robot pose!";
@@ -221,8 +224,8 @@ void MoveBaseAction::actionExePathFeedback(
     {
       std::stringstream oscillation_msgs;
       oscillation_msgs << "Robot is oscillating for " << (ros::Time::now() - last_oscillation_reset_).toSec() << "s!";
-      ROS_WARN_STREAM_NAMED("exe_path", oscillation_msgs.str());
-      action_client_exe_path_.cancelGoal();
+      ROS_WARN_STREAM_NAMED("move_base", oscillation_msgs.str());
+      cancelActiveClients();
 
       if (attemptRecovery())
       {
@@ -292,18 +295,18 @@ void MoveBaseAction::actionGetPathDone(
         ROS_WARN_STREAM_NAMED("move_base", "Abort the execution of the planner: " << get_path_result.message);
         goal_handle_.setAborted(move_base_result, state.getText());
       }
-      action_state_ =  FAILED;
+      action_state_ = FAILED;
       break;
 
     case actionlib::SimpleClientGoalState::PREEMPTED:
-      ROS_ASSERT_MSG(action_state_ == CANCELED,
-                     "GetPath PREEMPTED: action state must be CANCELED, not %d", action_state_);
-    ROS_WARN_STREAM_NAMED("move_base", "The action \"" << "get_path" << "\" was preempted successfully!  "<< action_state_ << "   must be "<< CANCELED);
-
-      // the get_path action has been preempted; copy result into move_base result
-      ROS_WARN_STREAM_NAMED("move_base", "The get_path action has been preempted!");
-      fillMoveBaseResult(get_path_result, move_base_result);
-      goal_handle_.setCanceled(move_base_result, state.getText());
+      ROS_INFO_STREAM_NAMED("move_base", "The action \"" << "get_path" << "\" has been preempted!");
+      if (action_state_ == CANCELED)
+      {
+        // move_base preempted while executing get_path; fill result and report canceled to the client
+        ROS_INFO_STREAM_NAMED("move_base", "move_base preempted while executing get_path");
+        fillMoveBaseResult(get_path_result, move_base_result);
+        goal_handle_.setCanceled(move_base_result, state.getText());
+      }
       break;
 
     case actionlib::SimpleClientGoalState::RECALLED:
@@ -355,7 +358,7 @@ void MoveBaseAction::actionExePathDone(
   // copy result from exe_path action
   fillMoveBaseResult(exe_path_result, move_base_result);
 
-  ROS_DEBUG_STREAM_NAMED("exe_path", "Current state:" << state.toString());
+  ROS_DEBUG_STREAM_NAMED("move_base", "Current state:" << state.toString());
 
   switch (state.state_)
   {
@@ -368,7 +371,7 @@ void MoveBaseAction::actionExePathDone(
       break;
 
     case actionlib::SimpleClientGoalState::ABORTED:
-      action_state_ =  FAILED;
+      action_state_ = FAILED;
 
       switch (exe_path_result.outcome)
       {
@@ -398,13 +401,14 @@ void MoveBaseAction::actionExePathDone(
       break;
 
     case actionlib::SimpleClientGoalState::PREEMPTED:
-      ROS_ASSERT_MSG(action_state_ == CANCELED,
-                     "ExePath PREEMPTED: action state must be CANCELED, not %d", action_state_);
-      ROS_WARN_STREAM_NAMED("move_base", "The action \"" << "exe_path" << "\" was preempted successfully!  "<< action_state_ << "   must be "<< CANCELED);
-
-      // the exe_path action has been preempted; copy result into move_base result
-      fillMoveBaseResult(exe_path_result, move_base_result);
-      goal_handle_.setCanceled(move_base_result, state.getText());
+      ROS_INFO_STREAM_NAMED("move_base", "The action \"" << "exe_path" << "\" has been preempted!");
+      if (action_state_ == CANCELED)
+      {
+        // move_base preempted while executing exe_path; fill result and report canceled to the client
+        ROS_INFO_STREAM_NAMED("move_base", "move_base preempted while executing exe_path");
+        fillMoveBaseResult(exe_path_result, move_base_result);
+        goal_handle_.setCanceled(move_base_result, state.getText());
+      }
       break;
 
     case actionlib::SimpleClientGoalState::RECALLED:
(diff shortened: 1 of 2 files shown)


mbf_abstract_nav/src/move_base_action.cpp
```diff
@@ -220,8 +224,8 @@ void MoveBaseAction::actionExePathFeedback(
     {
       std::stringstream oscillation_msgs;
       oscillation_msgs << "Robot is oscillating for " << (ros::Time::now() - last_oscillation_reset_).toSec() << "s!";
-      ROS_WARN_STREAM_NAMED("exe_path", oscillation_msgs.str());
-      action_client_exe_path_.cancelGoal();
+      ROS_WARN_STREAM_NAMED("move_base", oscillation_msgs.str());
+      cancelActiveClients();
```
spuetz: There is no other client running at that time, or am I wrong?

spuetz (commented):
A lot changes here, can you comment in the code why they are necessary?

mbf_abstract_nav/src/move_base_action.cpp
```diff
@@ -102,18 +102,22 @@ void MoveBaseAction::reconfigure(
 void MoveBaseAction::cancel()
 {
   action_state_ = CANCELED;
+  cancelActiveClients();
+}
 
-  if(!action_client_get_path_.getState().isDone())
+void MoveBaseAction::cancelActiveClients()
```
corot, line 108: Just extracting the cancelling into a single method cancelActiveClients

mbf_abstract_nav/src/move_base_action.cpp
```diff
@@ -148,9 +152,8 @@ void MoveBaseAction::start(GoalHandle &goal_handle)
   recovery_behaviors_ = goal.recovery_behaviors.empty() ? behaviors_ : goal.recovery_behaviors;
   current_recovery_behavior_ = recovery_behaviors_.begin();
 
-  geometry_msgs::PoseStamped robot_pose;
```
corot, line 151: No reason to create a local var.

mbf_abstract_nav/src/move_base_action.cpp
```diff
@@ -187,14 +190,15 @@ void MoveBaseAction::actionExePathActive()
 void MoveBaseAction::actionExePathFeedback(
     const mbf_msgs::ExePathFeedbackConstPtr &feedback)
 {
-  move_base_feedback_.outcome = feedback->outcome;
-  move_base_feedback_.message = feedback->message;
-  move_base_feedback_.angle_to_goal = feedback->angle_to_goal;
-  move_base_feedback_.dist_to_goal = feedback->dist_to_goal;
-  move_base_feedback_.current_pose = feedback->current_pose;
-  move_base_feedback_.last_cmd_vel = feedback->last_cmd_vel;
+  mbf_msgs::MoveBaseFeedback move_base_feedback;
```
corot, line 189: No reason to have a class attribute for feedback

mbf_abstract_nav/src/move_base_action.cpp
corot: True; reverted

mbf_abstract_nav/src/move_base_action.cpp
```diff
@@ -245,12 +249,7 @@ void MoveBaseAction::actionGetPathDone(
     const actionlib::SimpleClientGoalState &state,
     const mbf_msgs::GetPathResultConstPtr &result_ptr)
 {
-  if (action_state_ == CANCELED)
-    return;
```
corot, line 249: This return prevents to send the result

mbf_abstract_nav/src/move_base_action.cpp
```diff
@@ -356,24 +350,15 @@ void MoveBaseAction::actionExePathDone(
     const actionlib::SimpleClientGoalState &state,
     const mbf_msgs::ExePathResultConstPtr &result_ptr)
 {
-  if (action_state_ == CANCELED)
```
corot, line 359: This return prevents to send the result

mbf_abstract_nav/src/move_base_action.cpp
```diff
@@ -483,29 +474,22 @@ void MoveBaseAction::actionRecoveryDone(
   // give the robot some time to stop oscillating after executing the recovery behavior
   last_oscillation_reset_ = ros::Time::now();
 
-  if (action_state_ == CANCELED)
```
corot, line 486: This return prevents to send the result

corot:
> A lot changes here, can you comment in the code why they are necessary?

True, sorry. I added comments in the PR itself, explaining the changes. Non commented stuff is just basic refactoring.

mbf_abstract_nav/src/move_base_action.cpp
```diff
@@ -292,27 +291,22 @@ void MoveBaseAction::actionGetPathDone(
...
-
-      // copy result from get_path action
-      move_base_result.outcome = result.outcome;
-      move_base_result.message = result.message;
-      move_base_result.dist_to_goal = static_cast<float>(mbf_utility::distance(robot_pose_, goal_pose_));
-      move_base_result.angle_to_goal = static_cast<float>(mbf_utility::angle(robot_pose_, goal_pose_));
-      move_base_result.final_pose = robot_pose_;
-      goal_handle_.setCanceled(move_base_result, state.getText());
+      ROS_INFO_STREAM_NAMED("move_base", "The action \"" << "get_path" << "\" has been preempted!");
+      if (action_state_ == CANCELED)
```
corot, line 299: move base has been preempted. otherwise we have preempted internally, (what I don't think can happen) so we don't end the action

mbf_abstract_nav/src/move_base_action.cpp
```diff
@@ -406,18 +393,22 @@ void MoveBaseAction::actionExePathDone(
...
       }
       break;
 
     case actionlib::SimpleClientGoalState::PREEMPTED:
-      // action was preempted successfully!
-      ROS_DEBUG_STREAM_NAMED("move_base", "The action \""
-          << "exe_path" << "\" was preempted successfully!");
-      // TODO
+      ROS_INFO_STREAM_NAMED("move_base", "The action \"" << "exe_path" << "\" has been preempted!");
+      if (action_state_ == CANCELED)
```
corot, line 401: move base has been preempted. otherwise we have preempted internally, (what I don't think can happen) so we don't end the action

mbf_abstract_nav/src/move_base_action.cpp
```diff
@@ -540,10 +524,14 @@ void MoveBaseAction::actionRecoveryDone(
       );
       break;
     case actionlib::SimpleClientGoalState::PREEMPTED:
-      ROS_INFO_STREAM_NAMED("move_base",
-                             "The recovery action has been preempted!");
-      if(action_state_ == CANCELED)
-        goal_handle_.setCanceled();
+      ROS_INFO_STREAM_NAMED("move_base", "The action \"" << "recovery" << "\" has been preempted!");
+      if (action_state_ == CANCELED)
```
corot, line 524: move base has been preempted. otherwise we have preempted internally, (what I don't think can happen) so we don't end the action

mbf_abstract_nav/src/move_base_action.cpp
dorezyuk: i think checking for CANCELED is contra productive. 

We may end up at PREEMPTED by either calling SimpleActionClient::cancelGoal() or by someone else kicking us out of the current execution (see http://wiki.ros.org/actionlib/DetailedDescription)

In this class, we only call SimpleActionClient::cancelGoal() of action_client_get_path_ and action_client_recovery_ clients from MoveBaseAction::cancel(). So there is no "internal" preemption, and by removing this check we may react to another client stopping our goal. 

For action_client_exe_path_ its a bit more tricky, since we call SimpleActionClient::cancelGoal() on oscillation (so there is an internal preemption). What we might do is to reuse the recovery_trigger_ member to check inside the MoveBaseAction::actionExePathDone() if we want to react to a PREEMPTED flag or not:

```
void MoveBaseAction::actionExePathFeedback(
    const mbf_msgs::ExePathFeedbackConstPtr &feedback)
{
  // ...
  if (!oscillation_timeout_.isZero())
  {
    if (mbf_utility::distance(robot_pose_, last_oscillation_pose_) >= oscillation_distance_)
    {
    // ...
    else if (last_oscillation_reset_ + oscillation_timeout_ < ros::Time::now())
    {
      // ...
      // use this to indicate that we don't react to failures
      recovery_trigger_ = OSCILLATING;
      
      action_client_exe_path_.cancelGoal();

      if (!attemptRecovery())
      {
       // ...
      }
    }
  }
}

```
Then 
```
void MoveBaseAction::actionExePathDone(...){
// ...
switch(state.state_){
case PREEMPTED:
if(recovery_trigger_ != OSCILLATION) 
   goal_handle_.setCanceled(...);
}
}
```
And finally
```
void MoveBaseAction::actionRecoveryDone(...){
...
recovery_trigger_ = None;
}
```

corot: > i think checking for CANCELED is contra productive.
> 
> We may end up at PREEMPTED by either calling SimpleActionClient::cancelGoal() or by someone else kicking us out of the current execution (see http://wiki.ros.org/actionlib/DetailedDescription)

Don't think so, cause we only set action_state_ = CANCELED on cancel() method, and it only gets called we we cancel current move_base goal, not when preempting with a new one.

dorezyuk: my point is that if someone else preempts the current exec (by calling for example directly the exe_path action, instead of the move_base action), then the action_client_exe_path_ would signal PREEMPTED but the action_state_ would be not set to CANCELED. Then the client of the move_base action would hang

corot: I tried to do so using the ugly script below and it works, both master and this branch. Please give it a try and ideally modify it to trigger the cases you are worried about.

```
#!/usr/bin/env python

import time
import random

import rospy
import actionlib

from mag_common_py_libs.geometry import quaternion_from_yaw, pose2str

from mbf_msgs.msg import GetPathAction, GetPathGoal, GetPathResult
from mbf_msgs.msg import ExePathAction, ExePathGoal, ExePathResult
from mbf_msgs.msg import MoveBaseAction, MoveBaseGoal, MoveBaseResult

__author__ = 'santos'

""" 
Simplest possible MBF executive: sends goals provided through RViz on 2D Nav Goal tool to
move_base_flex/get_path action and sends the resulting path to move_base_flex/exe_path action
No fancy recovery or retry strategies.
"""


def create_geo_pose_stamped(x, y, yaw, frame_id='map'):
    pose = PoseStamped()
    pose.pose.position.x = x
    pose.pose.position.y = y
    pose.pose.orientation = quaternion_from_yaw(yaw)
    pose.header.frame_id = frame_id
    return pose

def nav_goal_cb(msg):
    global target_pose
    target_pose = msg
    goal = GetPathGoal(use_start_pose=False,
                       start_pose=msg,
                       target_pose=msg)
    get_path_ac.send_goal(goal, done_cb=plan_done_cb)
    rospy.loginfo("Relaying move_base_simple/goal %s from RViz to MBF's planner", pose2str(msg))
    time.sleep(1)
    goal = MoveBaseGoal(target_pose=msg,
                        controller='TEBPlanner')
    move_base_ac.send_goal(goal, done_cb=move_done_cb)

def plan_done_cb(status, result):
    if result.outcome == GetPathResult.SUCCESS:
        rospy.loginfo("Get path action succeeded")
        time.sleep(5)

        rospy.loginfo("Now Exe path !!!!!!!!!")
        goal = ExePathGoal(path=result.path,
                           controller='DWAPlanner')
        exe_path_ac.send_goal(goal, done_cb=follow_done_cb)
        target_pose = None
    else:
        rospy.logerr("Get path action failed with error code [%d]: %s", result.outcome, result.message)

def follow_done_cb(status, result):
    if result.outcome == GetPathResult.SUCCESS:
        rospy.loginfo("Follow path action succeeded")
    else:
        rospy.logerr("Follow path action failed with error code [%d]: %s", result.outcome, result.message)

def move_done_cb(status, result):
    if result.outcome == MoveBaseResult.SUCCESS:
        rospy.loginfo("Follow path action succeeded")
    else:
        rospy.logerr("Follow path action failed with error code [%d]: %s", result.outcome, result.message)

if __name__ == '__main__':
    rospy.init_node("simple_mbf_executive")

    target_pose = None

    get_path_ac = actionlib.SimpleActionClient("/move_base_flex/get_path", GetPathAction)
    get_path_ac.wait_for_server(rospy.Duration(5))
    exe_path_ac = actionlib.SimpleActionClient("/move_base_flex/exe_path", ExePathAction)
    exe_path_ac.wait_for_server(rospy.Duration(5))
    move_base_ac = actionlib.SimpleActionClient("/move_base_flex/move_base", MoveBaseAction)
    move_base_ac.wait_for_server(rospy.Duration(5))

    rospy.Subscriber('/move_base_simple/goal', PoseStamped, nav_goal_cb)

    rospy.spin()
```

spuetz (approved):

PR #219, opened 2020-08-20
Use `catkin_install_python` to install legacy relay. #noetic

FabianSchurig:
According to [Changing_shebangs](http://wiki.ros.org/UsingPython3/SourceCodeChanges#Changing_shebangs) we need to install python scripts with `catkin_install_python()` to let `catkin` decide whether to run it with Python 2 or 3. This makes it runnable on future distributions, e.g. Ubuntu 20.04 focal and ROS noetic, as well as on older ones like 18.04 bionic and melodic.

### DIFF
--- a/mbf_costmap_nav/CMakeLists.txt
+++ b/mbf_costmap_nav/CMakeLists.txt
@@ -105,6 +105,6 @@ install(DIRECTORY include/${PROJECT_NAME}/
   DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
 )
 
-install(PROGRAMS scripts/move_base_legacy_relay.py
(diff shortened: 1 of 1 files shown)


Rayman (approved):

corot:
Thanks @Bitfroest !

simonbogh:
Has this been released because I am still getting the error from binary install on Noetic?

rokusottervanger:
> Has this been released because I am still getting the error from binary install on Noetic?

I don't think so. I also got the error, so I compiled from source.

spuetz:
Nevertheless, I can release the current melodic and kinetic version for noetic. But noetic is developed in the noetic-devel branch.

PR #220, opened 2020-08-27
clear the costmap before deactivating it

dorezyuk:
Hi guys,

the clear_on_shutdown logic is flawed because obstacle- and voxel-grid-layers restart thier subscribers when their reset() is called. This means that the costmap itself is shutdown but the layers are still running in the background and processing the data.

I moved the reset before the stop. This is not ideal, since it is not guaranteed that the costmap will be really empty - but the best thing we can do without changing the costmap layers.

### DIFF
--- a/mbf_costmap_nav/src/mbf_costmap_nav/costmap_wrapper.cpp
+++ b/mbf_costmap_nav/src/mbf_costmap_nav/costmap_wrapper.cpp
@@ -130,10 +130,10 @@ void CostmapWrapper::deactivate(const ros::TimerEvent &event)
   boost::mutex::scoped_lock sl(check_costmap_mutex_);
 
   ROS_ASSERT_MSG(!costmap_users_, "Deactivating costmap with %d active users!", costmap_users_);
+  if (clear_on_shutdown_)
+    clear();  // do before stop, as some layers (e.g. obstacle and voxel) reactivate their subscribers on reset
   stop();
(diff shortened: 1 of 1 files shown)


corot:
Sorry, I don't get the point of this PR. To me sounds reasonable to stop the costmap and then clean it. If you clean it and then stop it, well, it can have the chance to go through another update round and undo the cleaning.

Anyway, I tried both codes and cannot see any difference in behavior. What's exactly the issue you are trying to fix? And how can I reproduce it?

dorezyuk:
When we shutdown the costmap, the layers deactivate their subscribers. In this manner they don't waste resources.

Our clear() implementation calls ultimately the reset() method of every loaded layer. The problem is, that the very popular `obstacle_layer` and `voxel_layer` plugins reactivate their subscribers on reset():

https://github.com/ros-planning/navigation/blob/c9fd50fb12265c653f4679d69300fce21c8b73ce/costmap_2d/plugins/obstacle_layer.cpp#L612

This leads to a relatively high idle load of the node, since the scan callbacks are relatively expensive.

For reproducing it
1. enable `obstacle_layer`, and configure `shutdown_costmaps` and `clear_on_shutdown` to true.
1. start the node and let it shutdown everything
1. `rosnode info mbf` should still indicate that the node still consumes the scan topics even after the shutdown

I hope this makes it clearer

corot:
Roger. Actually, I cannot see any difference on performance, but as you said, the topics are there!

PR #221, opened 2020-09-16
Fixed error: reference to TF is ambiguous

Navifra-Sally:
I have different TF class. So, ambiguous TF is fixed through the Scope Operator (::)


spuetz (approved):
lgtm

corot (approved):
lgtm

PR #231, opened 2020-10-28
Controller fails if robot pose gets older than tf_timeout

corot:
Current code uses the advanced tf2 interfaces, that returns a pose with stamp 0 (we request time 0 to get the latest available transform when getting robot pose). We then overwrite with time now, what is not realistic. Worse, if localization goes off, getRobotPose will keep providing the latest available tf without any warning. If we are not using costmaps, that means that the robot will keep moving while thinking he's standing!

This PR changes to the simple tf2 interfaces that returns the stamp of the latest available tf, and considers it as non valid if older than tf_timeout. Exe path will fail in that case, and the robot will stop, even if not using costmaps.
Moreover, using the simple tf2 interface allows us to simplify in turn the interface of the utility functions.

:warning: Not tested on Kinetic or older that use TF 1. Can anyone do it? :warning:

### DIFF
--- a/mbf_abstract_nav/src/abstract_controller_execution.cpp
+++ b/mbf_abstract_nav/src/abstract_controller_execution.cpp
@@ -169,14 +169,11 @@ std::vector<geometry_msgs::PoseStamped> AbstractControllerExecution::getNewPlan(
 
 bool AbstractControllerExecution::computeRobotPose()
 {
-  bool tf_success = mbf_utility::getRobotPose(*tf_listener_ptr, robot_frame_, global_frame_,
-                                              ros::Duration(tf_timeout_), robot_pose_);
-  // would be 0 if not, as we ask tf listener for the last pose available
-  robot_pose_.header.stamp = ros::Time::now();
-  if (!tf_success)
+  if (!mbf_utility::getRobotPose(*tf_listener_ptr, robot_frame_, global_frame_,
+                                 ros::Duration(tf_timeout_), robot_pose_))
   {
     ROS_ERROR_STREAM("Could not get the robot pose in the global frame. - robot frame: \""
-                         << robot_frame_ << "\"   global frame: \"" << global_frame_ << std::endl);
+                         << robot_frame_ << "\"   global frame: \"" << global_frame_);
     message_ = "Could not get the robot pose";
     outcome_ = mbf_msgs::ExePathResult::TF_ERROR;
     return false;
@@ -305,7 +302,8 @@ void AbstractControllerExecution::run()
 
       if (cancel_)
       {
-        if (force_stop_on_cancel_) {
+        if (force_stop_on_cancel_)
+        {
           publishZeroVelocity(); // command the robot to stop on canceling navigation
         }
         setState(CANCELED);
@@ -348,13 +346,21 @@ void AbstractControllerExecution::run()
       }
 
       // compute robot pose and store it in robot_pose_
-      computeRobotPose();
+      if (!computeRobotPose())
+      {
+        publishZeroVelocity();
+        setState(INTERNAL_ERROR);
+        condition_.notify_all();
+        moving_ = false;
+        return;
+      }
 
       // ask planner if the goal is reached
       if (reachedGoalCheck())
       {
         ROS_DEBUG_STREAM_NAMED("abstract_controller_execution", "Reached the goal!");
-        if (force_stop_at_goal_) {
+        if (force_stop_at_goal_)
+        {
           publishZeroVelocity();
         }
         setState(ARRIVED_GOAL);
--- a/mbf_abstract_nav/src/planner_action.cpp
+++ b/mbf_abstract_nav/src/planner_action.cpp
@@ -269,8 +269,8 @@ bool PlannerAction::transformPlanToGlobalFrame(
   for (iter = plan.begin(); iter != plan.end(); ++iter)
   {
     geometry_msgs::PoseStamped global_pose;
-    tf_success = mbf_utility::transformPose(robot_info_.getTransformListener(), robot_info_.getGlobalFrame(), iter->header.stamp,
-                                            robot_info_.getTfTimeout(), *iter, robot_info_.getGlobalFrame(), global_pose);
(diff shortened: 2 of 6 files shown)


mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -348,13 +346,21 @@ void AbstractControllerExecution::run()
       }
 
       // compute robot pose and store it in robot_pose_
-      computeRobotPose();
+      if (!computeRobotPose())
+      {
+        publishZeroVelocity();
```
corot, line 351: we should put all these lines in a setFailureState method...  I'll do in another PR

mbf_utility/include/mbf_utility/navigation_utility.h
```diff
@@ -58,38 +58,30 @@ namespace mbf_utility
...
  * @param tf_listener TransformListener.
  * @param target_frame Target frame for the point.
- * @param target_time Time, in that the frames should be used.
  * @param timeout Timeout for looking up the transformation.
  * @param in Point to transform.
- * @param fixed_frame Fixed frame of the source and target frame.
  * @param out Transformed point.
  * @return true, if the transformation succeeded.
  */
 bool transformPoint(const TF &tf,
```
corot, line 66: way cleaner interface

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -169,14 +169,11 @@ std::vector<geometry_msgs::PoseStamped> AbstractControllerExecution::getNewPlan(
 
 bool AbstractControllerExecution::computeRobotPose()
 {
-  bool tf_success = mbf_utility::getRobotPose(*tf_listener_ptr, robot_frame_, global_frame_,
-                                              ros::Duration(tf_timeout_), robot_pose_);
-  // would be 0 if not, as we ask tf listener for the last pose available
```
corot, line 174: Now contains the stamp of the latest available tf, much more accurate

mbf_abstract_nav/src/planner_action.cpp
```diff
@@ -269,8 +269,8 @@ bool PlannerAction::transformPlanToGlobalFrame(
   for (iter = plan.begin(); iter != plan.end(); ++iter)
   {
     geometry_msgs::PoseStamped global_pose;
-    tf_success = mbf_utility::transformPose(robot_info_.getTransformListener(), robot_info_.getGlobalFrame(), iter->header.stamp,
-                                            robot_info_.getTfTimeout(), *iter, robot_info_.getGlobalFrame(), global_pose);
```
dorezyuk, line 273: the changes in this file seems unrelated to the initial problem. To me both timestamps would be fine. Not sure if users might have specific applications running against the old implementation

mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
```diff
@@ -390,8 +390,7 @@ bool CostmapNavigationServer::callServiceCheckPointCost(mbf_msgs::CheckPoint::Re
   std::string costmap_frame = costmap->getGlobalFrameID();
 
   geometry_msgs::PointStamped point;
-  if (! mbf_utility::transformPoint(*tf_listener_ptr_, costmap_frame, request.point.header.stamp,
-                                     ros::Duration(0.5), request.point, global_frame_, point))
+  if (! mbf_utility::transformPoint(*tf_listener_ptr_, costmap_frame, ros::Duration(0.5), request.point, point))
```
dorezyuk: minor remark here and in the changes below: please fix also the formatting (white space after '!')

mbf_abstract_nav/src/planner_action.cpp
corot: Just adapt to the new, simpler interface of transformPose

mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
corot: which remark?   
format fixed

mbf_utility/include/mbf_utility/navigation_utility.h
spuetz: I would not say it's cleaner, but its simpler. But that's :+1:

mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
dorezyuk: the formatting was the remark ;)

mbf_abstract_nav/src/planner_action.cpp
dorezyuk: My point is that you change the behaviour from having a transform at request-timestamp to the newest transform. My concern is that this might brake existing user code. Since both implementations are reasonable, I'm inclined to keep the old one. @spuetz any preference here form your side?

corot: No, no;  we still get the newest transform. The only difference in behavior is that this transform will have the real time stamp, instead of time now. Actually, previous code was "lying"; with this PR we don't anymore

dorezyuk: Huh? I think you are talking about computeRobotPose/getRobotPose (which is a legit fix), and I'm talking transformPose and transformPoint: The old transformPose accepted target_time; the target_time was set to the time-stamp provided by the client (ex. line 272 in planner_action.cpp). Now transformPose uses the newest time.

corot: No, no; cause the new version uses the stamp provided with the pose; in this case, it's the same, as the pose is *iter and the stamp iter->header.stamp

dorezyuk: god you are right

dorezyuk (approved):

spuetz (approved):

PR #234, opened 2020-11-03
m&m for const-correctness

dorezyuk:
I've m&med all the mutexes from abstract_planner_execution and made all getters const; so it should be clearer that these functions don't alter the internal state of the class


corot (approved):
lgtm

corot:
Can you do the same for controller and recovery executions?

PR #235, opened 2020-11-03
avoid useless copying

dorezyuk:
Small changes to avoid useless copies for printing. Also add a comment to compute the cost_ (if its not provided by the planner) once instead of recomputing it every time getCost gets called (will do a dedicated PR for this)


mbf_abstract_nav/src/abstract_planner_execution.cpp
```diff
@@ -289,6 +289,7 @@ void AbstractPlannerExecution::run()
 
           plan_mtx_.lock();
           plan_ = plan;
+          // todo compute the cost once!
```
corot, line 292: You mean, recompute in getCost when cost_ is 0 and the plan doesn't change?
Yes, makes sense

dorezyuk: yep

PR #236, opened 2020-11-03
signal from setState function

dorezyuk:
I've moved the condition-variable inside setState, so we can easier trigger it on state-update. This simplifies the code a bit

### DIFF
--- a/mbf_abstract_nav/src/abstract_planner_execution.cpp
+++ b/mbf_abstract_nav/src/abstract_planner_execution.cpp
@@ -106,10 +106,14 @@ typename AbstractPlannerExecution::PlanningState AbstractPlannerExecution::getSt
   return state_;
 }
 
-void AbstractPlannerExecution::setState(PlanningState state)
+void AbstractPlannerExecution::setState(PlanningState state, bool signalling)
 {
   boost::lock_guard<boost::mutex> guard(state_mtx_);
   state_ = state;
+
+  // some states are quiet, most aren't
+  if(signalling)
+    condition_.notify_all();
 }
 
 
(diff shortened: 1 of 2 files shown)


corot (changes requested):
For controller, I had a different idea in mind: create a setFailedState method that joins these three lines:
```
          setState(<any error state>);
          publishZeroVelocity();
          condition_.notify_all();
          moving_ = false;
```
Does it make sense the same for planner?
In any case, whatever refactor we do, should be done also in recovery and controlling executions

dorezyuk:
I wouldn't mix the stuff and do the refactoring step by step. Right now there is no correspondent action to publishZeroVelocity for planners failing. We might later refactor the code into
```
setState(State _state, bool _signalling) {
...
planning_ = !_signalling
}
```
since we typically want to exit the planning loop if we signal

corot:
> I wouldn't mix the stuff and do the refactoring step by step. Right now there is no correspondent action to publishZeroVelocity for planners failing. We might later refactor the code into

As long as you follow up this PR with the matching changes for controller and recovery, I'm good approving this

corot (approved):

mbf_abstract_nav/src/abstract_planner_execution.cpp
```diff
@@ -106,10 +106,14 @@ typename AbstractPlannerExecution::PlanningState AbstractPlannerExecution::getSt
...
 }
 
-void AbstractPlannerExecution::setState(PlanningState state)
+void AbstractPlannerExecution::setState(PlanningState state, bool signalling)
 {
   boost::lock_guard<boost::mutex> guard(state_mtx_);
   state_ = state;
+
+  // some states are quiet, most aren't
+  if(signalling)
```
corot, line 115: missing space after if

dorezyuk: would it make sense to add a clang-format file so we use common formatting?

corot: yes,,,  #239

spuetz (approved):

PR #237, opened 2020-11-03
Dima/planner

dorezyuk:
I've refactored/simplified the implementation of AbstractPlannerExecution::run and added tests for the method. See comments in the code for explanation

### DIFF
--- a/mbf_abstract_nav/CMakeLists.txt
+++ b/mbf_abstract_nav/CMakeLists.txt
@@ -96,11 +96,8 @@ if(CATKIN_ENABLE_TESTING)
     test/abstract_action_base.cpp)
   target_link_libraries(abstract_action_base_test ${MBF_ABSTRACT_SERVER_LIB})
   
-  # gmock-tests only for noetic and newer
-  if(${roscpp_VERSION} VERSION_GREATER 1.15.0)
-    add_rostest_gmock(abstract_planner_execution_test
-      test/abstract_planner_execution.launch
-      test/abstract_planner_execution.cpp)
-    target_link_libraries(abstract_planner_execution_test ${MBF_ABSTRACT_SERVER_LIB})
-  endif()
+  add_rostest_gmock(abstract_planner_execution_test
+    test/abstract_planner_execution.launch
+    test/abstract_planner_execution.cpp)
+  target_link_libraries(abstract_planner_execution_test ${MBF_ABSTRACT_SERVER_LIB})
 endif()
--- a/mbf_abstract_nav/CMakeLists.txt
+++ b/mbf_abstract_nav/CMakeLists.txt
@@ -90,8 +90,17 @@ install(DIRECTORY include/${PROJECT_NAME}/
 if(CATKIN_ENABLE_TESTING)
   find_package(rostest REQUIRED)
 
+  # ros-tests
   add_rostest_gmock(abstract_action_base_test
     test/abstract_action_base.launch
     test/abstract_action_base.cpp)
   target_link_libraries(abstract_action_base_test ${MBF_ABSTRACT_SERVER_LIB})
+  
+  # gmock-tests only for noetic and newer
+  if(${roscpp_VERSION} VERSION_GREATER 1.15.0)
+    add_rostest_gmock(abstract_planner_execution_test
+      test/abstract_planner_execution.launch
+      test/abstract_planner_execution.cpp)
+    target_link_libraries(abstract_planner_execution_test ${MBF_ABSTRACT_SERVER_LIB})
+  endif()
 endif()
--- a/mbf_abstract_nav/src/abstract_planner_execution.cpp
+++ b/mbf_abstract_nav/src/abstract_planner_execution.cpp
@@ -226,10 +226,6 @@ void AbstractPlannerExecution::run()
   geometry_msgs::PoseStamped current_goal = goal_;
   double current_tolerance = tolerance_;
 
-  bool success = false;
-  bool make_plan = false;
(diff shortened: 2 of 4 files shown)


mbf_abstract_nav/src/abstract_planner_execution.cpp
```diff
@@ -312,43 +308,33 @@ void AbstractPlannerExecution::run()
...
         {
           ROS_INFO_STREAM("Planning could not find a plan!");
-          exceeded = true;
-          planning_ = false;
           setState(NO_PLAN_FOUND, true);
         }
         else
         {
-          exceeded = false;
+          planning_ = true;
```
dorezyuk: we used to set planning_ = false in all cases but this. Now I set it once to false and change it here to true to reduce verbosity

mbf_abstract_nav/src/abstract_planner_execution.cpp
```diff
@@ -261,47 +256,48 @@ void AbstractPlannerExecution::run()
...
         ROS_INFO_STREAM("New goal pose: (" << g.x << ", " << g.y << ", " << g.z << ")");
       }
 
-      make_plan = !(success || exceeded) || has_new_start_ || has_new_goal_;
-
       // unlock goal
       goal_start_mtx_.unlock();
-      setState(PLANNING, false);
-      if (make_plan)
+      if (cancel_)
```
dorezyuk, line 269: moved the cancel up so we don't call the costly makePlan method, if the user has already cancelled us

mbf_abstract_nav/src/abstract_planner_execution.cpp
```diff
@@ -261,47 +256,48 @@ void AbstractPlannerExecution::run()
         current_tolerance = tolerance_;
         ROS_INFO_STREAM("A new goal pose is available. Planning with the new goal pose and the tolerance: "
                         << current_tolerance);
-        exceeded = false;
         geometry_msgs::Point g = goal_.pose.position;
         ROS_INFO_STREAM("New goal pose: (" << g.x << ", " << g.y << ", " << g.z << ")");
       }
 
-      make_plan = !(success || exceeded) || has_new_start_ || has_new_goal_;
```
dorezyuk, line 269: make_plan is actually redundant, since we would set planning_ to false in all but the 'retry' case. Also in the old implementation we had an unhandled else case (which could never happen, but still) if which cancel_ and make_plan would be false. In this case we would go for busy spinning at full speed

mbf_abstract_nav/src/abstract_planner_execution.cpp
```diff
@@ -261,47 +256,48 @@ void AbstractPlannerExecution::run()
...
-          plan_mtx_.lock();
+          boost::lock_guard<boost::mutex> plan_mtx_guard(plan_mtx_);
           plan_ = plan;
           cost_ = cost;
           last_valid_plan_time_ = ros::Time::now();
-          plan_mtx_.unlock();
           setState(FOUND_PLAN, true);
         }
-        else if (max_retries_ >= 0 && ++retries > max_retries_)
+        else if (max_retries_ > 0 && ++retries > max_retries_)
```
dorezyuk, line 302: mind here the strict greater. With the >= case, we can never enter NO_PLAN_FOUND case

mbf_abstract_nav/src/abstract_planner_execution.cpp
```diff
@@ -312,43 +308,33 @@ void AbstractPlannerExecution::run()
...
     ROS_WARN_STREAM("Planner thread interrupted!");
-    planning_ = false;
     setState(STOPPED, true);
   }
   catch (...)
   {
     ROS_FATAL_STREAM("Unknown error occurred: " << boost::current_exception_diagnostic_information());
     setState(INTERNAL_ERROR, true);
   }
+  planning_ = false;
```
dorezyuk: if the planner would throw, we would miss to reset the planning_ to false (then start() would always return false)

mbf_abstract_nav/CMakeLists.txt
```diff
@@ -87,3 +87,15 @@ install(DIRECTORY include/${PROJECT_NAME}/
   DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
 )
 
+if(CATKIN_ENABLE_TESTING)
+  find_package(rostest REQUIRED)
+
+  # ros-tests
+  # gmock-tests only for noetic and newer
+  if(${roscpp_VERSION} VERSION_GREATER 1.15.0)
```
dorezyuk: for now only noetic...

corot (changes requested):
I got an error while running the tests:

image

dorezyuk:
> I got an error while running the tests:
> 
> image

Thats part of the test - have a look at the "exception" test-case

corot:
> Thats part of the test - have a look at the "exception" test-case

Wonder if we can make the exception quiet,,,,  :thinking: 

Apart from that, lgtm

corot (approved):

dorezyuk:
> Wonder if we can make the exception quiet,,,, thinking

Not sure if I understand: do you mean within move-base-flex or within the test?

corot:
Within the test;  check that MBF throws the exception w/o showing on the
test results

not a critical thing,,,  just aesthetics (the error looks like
contradicting all the PASSED msgs)

dorezyuk:
I've changed the log-level of the test to fatal and changed the print to error - now it should look better

spuetz (approved):

PR #239, opened 2020-11-04
Add clang format and do not git-ignore it

corot:
Based on https://github.com/davetcoleman/roscpp_code_format/blob/master/.clang-format

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,65 @@
+---
+BasedOnStyle: Google
+AccessModifierOffset: -2
+ConstructorInitializerIndentWidth: 2
+AlignEscapedNewlinesLeft: false
+AlignTrailingComments: true
+AllowAllParametersOfDeclarationOnNextLine: false
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: None
+AlwaysBreakTemplateDeclarations: true
+AlwaysBreakBeforeMultilineStrings: false
+BreakBeforeBinaryOperators: false
+BreakBeforeTernaryOperators: false
+BreakConstructorInitializersBeforeComma: true
+BinPackParameters: true
+ColumnLimit: 120
+ConstructorInitializerAllOnOneLineOrOnePerLine: true
+DerivePointerBinding: false
+PointerBindsToType: true
+ExperimentalAutoDetectBinPacking: false
+IndentCaseLabels: true
+MaxEmptyLinesToKeep: 1
+NamespaceIndentation: None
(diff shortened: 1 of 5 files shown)


dorezyuk (approved):

Rayman:
Are you planning to `clang-format` the full repository, or is it already formatted?

corot:
> Are you planning to `clang-format` the full repository, or is it already formatted?

mmm....  :thinking:   not really;  code is mostly formatted and I don't want to destroy manually aligned code. But I can highlight format errors on CLion.

dorezyuk:
if you want to exclude stuff from being formatted by clang then wrap it in
```
// clang-format off
.. bla
// clang-format on
```

corot:
Shit... I should have done this after merging open PRs... :frowning:

spuetz (changes requested):
This destroys the git history for individual lines. This is not good. We should have a format checker only for PRs (hich also just checks the changes). Please do not reformat the whole repo. I am absolutely against that.

corot:
> This destroys the git history for individual lines. This is not good. We should have a format checker only for PRs. Please do not reformat the whole repo. I am absolutely against that.

True, true... but ignoring white-spaces the history is still recognizable, don't you think?
Is there any way to force git to ALWAYS ignore white-spaces?

Rayman:
You can configure the CI to only verify the format of the current pull request. That way the code get's incrementally better formatted. See `clang-format-diff`

corot:
I reverted the formatting;  now just add the clang format file, so we can configure CI to check if new PRs adhere to it

spuetz:
Could you also change the CI script to use https://clang.llvm.org/docs/ClangFormat.html, e.g. like this here: https://gitlab.gnome.org/GNOME/glib/-/blob/0ea992a9d8378189868c27c181ee00e390f5f346/.gitlab-ci/run-style-check-diff.sh and here: https://gitlab.gnome.org/GNOME/glib/blob/0ea992a9d8378189868c27c181ee00e390f5f346/.gitlab-ci.yml#L33-39

corot:
> Could you also change the CI script to use https://clang.llvm.org/docs/ClangFormat.html, e.g. like this here: https://gitlab.gnome.org/GNOME/glib/-/blob/0ea992a9d8378189868c27c181ee00e390f5f346/.gitlab-ci/run-style-check-diff.sh and here: https://gitlab.gnome.org/GNOME/glib/blob/0ea992a9d8378189868c27c181ee00e390f5f346/.gitlab-ci.yml#L33-39

mmm...  I added both scripts to the PR and run the first on .travis.yml, but I have no idea how to test if this will work as we expect. Any help will be welcome

PR #242, opened 2020-11-12
Dima/abstract action base

dorezyuk:
Hey guys,

in this PR I fixup some minor issues and provide basic testing to the abstract_action_base class.

List of changes

## Misc
1. add missing includes
2. add some documentation for the abstract_action_base class
3. add a typedef for the concurrency map; we should change this later to unordered_map
4. avoid costly calls to concurrency_slots_[slot] but do it once

## API adjustments
1. this is really debatable but I added a runImpl function, which allows us to skip the boost::bind stuff. i think it expresses/documents  better what is exactly required to implement the action base
2. We call execution.reconfigure() from the abstract_action_base. I've added that as a virtual method to the abstract_execution_base to account for that.

## Fixes
1. Our destructor would not join the threads. fixed that.

### DIFF
--- a/mbf_abstract_nav/CMakeLists.txt
+++ b/mbf_abstract_nav/CMakeLists.txt
@@ -90,8 +90,8 @@ install(DIRECTORY include/${PROJECT_NAME}/
 if(CATKIN_ENABLE_TESTING)
   find_package(rostest REQUIRED)
 
-  catkin_add_gmock(${PROJECT_NAME}_test test/abstract_action_base.cpp)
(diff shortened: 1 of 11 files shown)


corot:
And can you adopt the clang format from https://github.com/magazino/move_base_flex/pull/239?

dorezyuk:
I know sorry for the spamming

spuetz (approved):
Looks very good! :+1:

corot (approved):
Great to start having at least some test. Thanks!
A couple of comments, specially renaming test launcher to .test

PR #243, opened 2020-11-24
fix blind driving

dorezyuk:
Hey guys,

In the PR #198 we changed the signature of the abstract_navigation_server. However, the signature of the costmap_navigation_server was forgotten; the effect was that we were driving "blindly" - without a costmap. This PR fixes the issue. Maybe we can at one point embrace c++11 (override would have helped here)


Rayman:
I think we should be using c++11. Kinetic supports that so we should be fine: https://www.ros.org/reps/rep-0003.html#kinetic-kame-may-2016-may-2021

PR #248, opened 2020-12-22
Dima/add clang format

dorezyuk:
Hi guys,

I've took the pr from Jorge (#239) and adjusted it so it runs on the github/workflows ci.

The behaviour is:
- runs only on pull-requests
- applies the clang-format and checks if there would be a diff. If there is a diff (== the pr has ill-formatted code), it fails.

Best,
Dima


corot:
Thanks!!! :smile:

PR #249, opened 2020-12-23
Dima/abstract execution base

dorezyuk:
Hi guys, 

The testing goes on. In this pr its about the abstract_execution_base.

# MInor adjustments
- pass and get strings by const ref instread of by value

# Major Fixes
- we never joined threads when destructing the class.
- we never checked if the thread_ actually was stopped when starting it.
- I get a segfault if the instance would be destructed right after calling start(). The exact reason for this is rather complex (== not sure if I understand it fully myself) but converting the pure virtual methods to just virtual methods solves it.
- we never reset cancel_ and other execution related data. As far as i can see this would render the whole class unusable if the user would cancel an action once.

Best,
Dima

### DIFF
--- a/mbf_abstract_nav/include/mbf_abstract_nav/abstract_execution_base.h
+++ b/mbf_abstract_nav/include/mbf_abstract_nav/abstract_execution_base.h
@@ -71,7 +71,10 @@ class AbstractExecutionBase
     * @brief Cancel the plugin execution.
     * @return true, if the plugin tries / tried to cancel the computation.
     */
-   virtual bool cancel() = 0;
+   virtual bool cancel()
+   {
+     return false;
+   };
 
    void join();
 
@@ -109,7 +112,7 @@ class AbstractExecutionBase
    {}
 
 protected:
-  virtual void run() = 0;
+  virtual void run(){};
 
   //! condition variable to wake up control thread
   boost::condition_variable condition_;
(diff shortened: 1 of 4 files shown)


mbf_abstract_nav/include/mbf_abstract_nav/abstract_execution_base.h
```diff
@@ -45,61 +45,75 @@
...
+ */
 class AbstractExecutionBase
 {
  public:
+   AbstractExecutionBase(const std::string& name);
 
-  AbstractExecutionBase(std::string name);
+   virtual ~AbstractExecutionBase();
 
-  virtual bool start();
```
corot, line 57: The indentation is 2 spaces, so it was correct

corot (changes requested):
Indentation is 2 spaces (this changes to 3 on mbf_abstract_nav/include/mbf_abstract_nav/abstract_execution_base.h).
Other than that, lgtm
Thanks!

mbf_abstract_nav/include/mbf_abstract_nav/abstract_execution_base.h
dorezyuk: Yep, that definitely odd, but I fear this is an artefact of the "partial" formatting we apply. If i edit it manually, our CI will fail. The only thing I can think of is to reformat the whole file - and im not sure if its really worth it

corot: No idea;  anyway, I formatted locally

corot: lgtm

corot (approved):

corot:
@dorezyuk, can u rebase with latest master?

dorezyuk:
rebased

spuetz (approved):

PR #254, opened 2021-01-20
fix race-condition in the test

dorezyuk:


corot (approved):
never experience this...  can I reproduce somehow?

dorezyuk:
I think thats a race condition in my test - should be fine now

PR #256, opened 2021-01-21
Dima/transform pose

dorezyuk:
Hey, 
I've added a small patch to transform the given poses into the global frame id of the global costmap - this is more or less an implicit requirement for many global planners.

### DIFF
--- a/mbf_costmap_nav/src/mbf_costmap_nav/costmap_planner_execution.cpp
+++ b/mbf_costmap_nav/src/mbf_costmap_nav/costmap_planner_execution.cpp
@@ -44,15 +44,11 @@
 
 namespace mbf_costmap_nav
 {
-
-CostmapPlannerExecution::CostmapPlannerExecution(
-    const std::string &planner_name,
-    const mbf_costmap_core::CostmapPlanner::Ptr &planner_ptr,
-    const TFPtr &tf_listener_ptr,
-    const CostmapWrapper::Ptr &costmap_ptr,
-    const MoveBaseFlexConfig &config)
-      : AbstractPlannerExecution(planner_name, planner_ptr, tf_listener_ptr, toAbstract(config)),
-        costmap_ptr_(costmap_ptr)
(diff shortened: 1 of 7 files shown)


dorezyuk:
this kinetic stuff is really painful

mbf_costmap_nav/src/mbf_costmap_nav/costmap_planner_execution.cpp
```diff
@@ -77,12 +75,26 @@ uint32_t CostmapPlannerExecution::makePlan(const geometry_msgs::PoseStamped &sta
                                            double &cost,
                                            std::string &message)
 {
+  // transform the input to the global frame of the costmap, since this is an
+  // "implicit" requirement for most planners
+  // note: costmap_2d::Costmap2DROS::getTransformTolerance might be a good idea,
+  // but it's not part of the class API in ros-kinetic
+  const ros::Duration pat(0.5);
```
corot: can we pass directly the values to transformPose?  or at least make frame a reference

corot (approved):
I have never faced the need of this, but sounds reasonable. I suppose most planners do the conversion themselves, e,g, [global_planner check the frames](https://github.com/ros-planning/navigation/blob/noetic-devel/global_planner/src/planner_core.cpp#L234)

The fact that u can provide start/goal poses in any reference frame should be documented somehow,,, maybe in GetPath.action?

mbf_costmap_nav/src/mbf_costmap_nav/costmap_planner_execution.cpp
dorezyuk: the getGlobalFrameID returns by value for kinetic and melodic, so we cannot use a ref

corot: Ack;  can we rename pat as timeout?

spuetz (approved):

Issue #167, opened 2020-01-08
controller updates in "real"-time in simulation


MCFurry:
While running mbf in a simulation where the simulation doesn't reach (or is not set to) a 1.0 real-time factor, mbf still calls controller execution in "real"-time since it uses the boost chrono timer:
[link](https://github.com/magazino/move_base_flex/blob/e4cdbb9f9ca0191cfd1905ff5125f67685292534/mbf_abstract_nav/src/abstract_controller_execution.cpp#L375)
This way the performance of the controller largely depends on how fast we (can) run a simulation and thus yields different results, or even problems in some scenarios, since the ROS time is not synchronous with the execution time of the controller.

Is this intended behaviour? Or how would you advise to run simulations with mbf?

spuetz:
You're right! Thank you for mentioning this error. Honestly, I didn't think about it.

MCFurry:
So maybe switching the timing to ROS-based timing would be the solution?
Would this in your opinion be "the best" solution?

corot:
I suppose using [ROS Timers](http://wiki.ros.org/roscpp/Overview/Timers) will do the trick.
PR welcome!!!

corot:
fixed on master

PR #257, opened 2021-01-26
Use ROS rates for sleeping to support simulation time

Timple:
Solves #167

At the moment non-functional. But if you like the approach and don't see any obstacles I'll finish this implementation.

### DIFF
--- a/mbf_abstract_nav/src/abstract_controller_execution.cpp
+++ b/mbf_abstract_nav/src/abstract_controller_execution.cpp
@@ -422,7 +422,7 @@ void AbstractControllerExecution::run()
         boost::this_thread::interruption_point();
         if (!loop_rate_.sleep())
         {
-          ROS_WARN_THROTTLE(1.0, "Calculation needs too much time to stay in the moving frequency! (%f > %f)",
+          ROS_WARN_THROTTLE(1.0, "Calculation needs too much time to stay in the moving frequency! (%.4fs > %.4fs)",
                             loop_rate_.cycleTime().toSec(), loop_rate_.expectedCycleTime().toSec());
         }
         boost::this_thread::interruption_point();
--- a/mbf_abstract_nav/src/abstract_controller_execution.cpp
+++ b/mbf_abstract_nav/src/abstract_controller_execution.cpp
@@ -418,10 +418,10 @@ void AbstractControllerExecution::run()
 
       if (moving_ && ros::ok())
       {
+        // provide an interruption point also with 0 or negative sleep_time
+        boost::this_thread::interruption_point();
         if (!loop_rate_.sleep())
         {
-          // provide an interruption point also with 0 or negative sleep_time
-          boost::this_thread::interruption_point();
           ROS_WARN_THROTTLE(1.0, "Calculation needs too much time to stay in the moving frequency! (%f > %f)",
                             loop_rate_.cycleTime().toSec(), loop_rate_.expectedCycleTime().toSec());
         }
--- a/mbf_abstract_nav/src/abstract_controller_execution.cpp
+++ b/mbf_abstract_nav/src/abstract_controller_execution.cpp
@@ -57,7 +57,7 @@ AbstractControllerExecution::AbstractControllerExecution(
   AbstractExecutionBase(name),
     controller_(controller_ptr), tf_listener_ptr(tf_listener_ptr), state_(INITIALIZED),
     moving_(false), max_retries_(0), patience_(0), vel_pub_(vel_pub), current_goal_pub_(goal_pub),
-    loop_rate_(ros::Rate(DEFAULT_CONTROLLER_FREQUENCY))
+    loop_rate_(DEFAULT_CONTROLLER_FREQUENCY)
 {
   ros::NodeHandle nh;
   ros::NodeHandle private_nh("~");
--- a/mbf_abstract_nav/src/abstract_controller_execution.cpp
+++ b/mbf_abstract_nav/src/abstract_controller_execution.cpp
@@ -57,7 +57,7 @@ AbstractControllerExecution::AbstractControllerExecution(
   AbstractExecutionBase(name),
     controller_(controller_ptr), tf_listener_ptr(tf_listener_ptr), state_(INITIALIZED),
     moving_(false), max_retries_(0), patience_(0), vel_pub_(vel_pub), current_goal_pub_(goal_pub),
-    calling_duration_(boost::chrono::microseconds(static_cast<int>(1e6 / DEFAULT_CONTROLLER_FREQUENCY)))
+    loop_rate_(ros::Rate(DEFAULT_CONTROLLER_FREQUENCY))
 {
   ros::NodeHandle nh;
   ros::NodeHandle private_nh("~");
@@ -88,7 +88,7 @@ bool AbstractControllerExecution::setControllerFrequency(double frequency)
     ROS_ERROR("Controller frequency must be greater than 0.0! No change of the frequency!");
     return false;
   }
-  calling_duration_ = boost::chrono::microseconds(static_cast<int>(1e6 / frequency));
+  loop_rate_ = ros::Rate(frequency);
   return true;
 }
 
@@ -298,8 +298,6 @@ void AbstractControllerExecution::run()
   {
     while (moving_ && ros::ok())
     {
-      boost::chrono::thread_clock::time_point loop_start_time = boost::chrono::thread_clock::now();
-
       if (cancel_)
       {
         if (force_stop_on_cancel_)
@@ -317,7 +315,7 @@ void AbstractControllerExecution::run()
         // the specific implementation must have detected a risk situation; at this abstract level, we
         // cannot tell what the problem is, but anyway we command the robot to stop to avoid crashes
         publishZeroVelocity();   // note that we still feedback command calculated by the plugin
-        boost::this_thread::sleep_for(calling_duration_);
+        loop_rate_.sleep();
       }
 
       // update plan dynamically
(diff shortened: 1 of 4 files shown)


corot:
Yup, make sense to use ROS Rate, as long as we can interrupt the sleep stopping the thread (currently all calls to boost::this_thread::sleep_for are potential interruption points).

Timple:
Under the hood, ROS sleeps use nanosleep. Not sure the interrupting part works the same.

But the sleep is the last statement in the loop. So no code is executed in case the sleep is not interrupted. And the next cycle of course catches the interruption. 
Worst-case this only ads a slight delay of once cycle of doing nothing right?

corot:
yes,,, but that's already a bad thing  :thinking:
can u make a fast experiment to see if sleep gets interrupted?

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -418,25 +416,14 @@ void AbstractControllerExecution::run()
...
-        if (sleep_time > boost::chrono::microseconds(0))
-        {
-          // interruption point
-          boost::this_thread::sleep_for(sleep_time);
-        }
-        else
+        if (!loop_rate_.sleep())
         {
           // provide an interruption point also with 0 or negative sleep_time
           boost::this_thread::interruption_point();
```
corot: If loop_rate_.sleep() doesn't honor interruptions, we should move this out of the if

Timple:
> can u make a fast experiment to see if sleep gets interrupted?

http://coliru.stacked-crooked.com/a/a8fc934550deca50

The nanosleeps that ROS uses are also interrupted :slightly_smiling_face:

dorezyuk:
> > can u make a fast experiment to see if sleep gets interrupted?
> 
> http://coliru.stacked-crooked.com/a/a8fc934550deca50
> 
> The nanosleeps that ROS uses are also interrupted slightly_smiling_face

I think the code you've posted shows that nanosleep is **not** an interruption point. To make it better visible you should take the end_time after calling join and increase the sleep-duration passed to nanosleep.

Also this discussion on [stackoverflow](https://stackoverflow.com/questions/3072912/boostthis-threadsleep-vs-nanosleep) suggests that nanosleep is not an interruption point

dorezyuk:
```

int main() {

    boost::thread t([]{
		try {
			// boost::this_thread::sleep_for(boost::chrono::milliseconds(500));
			nanosleep((const struct timespec[]){{2, 500000L}}, NULL);
		}
		
		catch (...) { std::cout << "Interrupted\n" << std::endl; }
	});
	auto start_time = boost::chrono::steady_clock::now();
	boost::this_thread::sleep_for(boost::chrono::milliseconds(100));
	t.interrupt();
	t.join();
	auto end_time = boost::chrono::steady_clock::now();
	std::cout << "Ran for: " << boost::chrono::duration_cast<boost::chrono::seconds>(end_time - start_time)<< "\n";
}
```
Returns `Ran for: 2 seconds`

Timple:
Ah, obviously I needed to put the `end_time` after `t.join()`. My bad.

---

So the ROS time doesn't interrupt. Would it make sense to keep the original sleep logic (albeit a duplicate of [ros::rate](http://docs.ros.org/en/diamondback/api/rostime/html/rate_8cpp_source.html#l00040)). And replace the instances of `boost::chrono::thread_clock::now();` by `ros::Time::now();`?

spuetz:
> Would it make sense to keep the original sleep logic (albeit a duplicate of ros::rate). And replace the instances of boost::chrono::thread_clock::now(); by ros::Time::now();?

Sounds reasonable. You could also just use the internal time value from `ros::Time::now()`, or?

Timple:
`ros::Time::now()` already adheres to walltime or simulation time depending on what's active.

dorezyuk:
> > Would it make sense to keep the original sleep logic (albeit a duplicate of ros::rate). And replace the instances of boost::chrono::thread_clock::now(); by ros::Time::now();?
> 
> Sounds reasonable. You could also just use the internal time value from `ros::Time::now()`, or?

I think swapping boost::system_clock for ros::Time will not work, since boost::this_thread::sleep_for will always sleep in system time. You would need then to scale the sleep_time somehow.

We might do something ugly like
```cpp
void mbf_sleep(const ros::Duration& _dur) {
 const ros::Time end = ros::Time::now() + _dur;
 while(ros::Time::now() < end) {
    // this is now an interruption point
    boost::this_thread::sleep_for(boost::chrono::milliseconds(10));
 }
}
```
But thats very close to busy waiting

spuetz:
It should be possible to compute the time delta and estimate the sleep duration in the time system of choice.

Timple:
What goes wrong if we do not interrupt?

1. The publishZero() is delayed by max 1/control_freq

Anything else?

spuetz:
The question should be what goes wrong if we interrupt and nothings happens, as in the most cases anyway, since there are most likely no interruption points in the respective controller plugins and the duration after executing the plugin to to the end of the loop is negligible short. The important thing here is more that we could interrupt plugins / beside cancelling, meaning that there could be interruption points in the plugin. So from my point of view it is fine if the sleep duration is small enough.

dorezyuk:
> It should be possible to compute the time delta and estimate the sleep duration in the time system of choice.

The issue I see here is that the time-scaling may vary from loop to loop. Imagine you run Gazebo and press pause for example. In this particular case the time "freezes" (which would make us sleep for a long time) and its impossible to predict when it will resume ticking

spuetz:
> The issue I see here is that the time-scaling may vary from loop to loop. Imagine you run Gazebo and press pause for example. In this particular case the time "freezes" (which would make us sleep for a long time) and its impossible to predict when it will resume ticking

Ok, sure the pause is a special case, but computing the sleep duration assuming a linear time scaling should be doable. The pause could be detected by a time delta which equals zero, right?

Anyway. I suggest to ignore the interruption here, as currently it is unused in all cases and the more important thing is interrupting plugins, which is still possible.

spuetz (approved):

spuetz (commented):
Is it functional in the current state? Should be, or?

Timple:
It was functional in simulation before the last commit. I assume it still is.

I can run some tests tomorrow!

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -418,25 +416,14 @@ void AbstractControllerExecution::run()
...
-      configuration_mutex_.unlock();
       if (moving_ && ros::ok())
       {
-        if (sleep_time > boost::chrono::microseconds(0))
-        {
-          // interruption point
-          boost::this_thread::sleep_for(sleep_time);
-        }
-        else
+        // provide an interruption point also with 0 or negative sleep_time
```
corot: the comment should be now something like "provide an interruption point, as the nanosleep used by ROS Timer is not  interruptable"

corot: Would it make sense to add another interruption_point after the sleep?  :thinking:

Timple: Maybe at the start of the loop makes more sense? On case one interrupts immediately

Timple: Although before and after sleep might be more explicit because that's the actual reason

corot: yap,,,  taht's what I thought;  but looks a bit... shoddy, hehe.   I'm good with either option;   will give a try

spuetz (approved):
lgtm.

Timple:
The dynamic reconfigure of the frequency seems to be broken. When I change it from 20hz to 100hz I get the following message:

```
Calculation needs too much time to stay in the moving frequency! (0.0400s > 0.0100s)
```

So somehow the rate doesn't change but the reported expected rate does...
I tried re-initialising the class with `std::make_unique` but that results in the same behavior. Any suggestions here?

spuetz:
Hmmm 20Hz should be 0.05sec. How often does the message show up?  Line 91 should to the trick `loop_rate_ = ros::Rate(frequency);` I'm wondering why it does not.

So maybe 0.04 sec is what the plugin needs?

Timple:
Ah yes, that was it. I thought it was exactly the frequency.

0.04 is exactly the discretisation step used in my current simulation. So obviously everything takes at least that long...

I'm good, it's working now!

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -57,7 +57,7 @@ AbstractControllerExecution::AbstractControllerExecution(
   AbstractExecutionBase(name),
     controller_(controller_ptr), tf_listener_ptr(tf_listener_ptr), state_(INITIALIZED),
     moving_(false), max_retries_(0), patience_(0), vel_pub_(vel_pub), current_goal_pub_(goal_pub),
-    calling_duration_(boost::chrono::microseconds(static_cast<int>(1e6 / DEFAULT_CONTROLLER_FREQUENCY)))
+    loop_rate_(ros::Rate(DEFAULT_CONTROLLER_FREQUENCY))
```
corot: slightly simpler: `loop_rate_(DEFAULT_CONTROLLER_FREQUENCY)`

corot (approved):
Minor change on loop_rate_ initialization;  apart from that, lgtm
Tested with stage changing real time rate up and down and works fine

corot:
thanks! :+1:

Timple:
Thank you for the fast review cycles. Didn't expect to get so much traction :+1:

Timple:
All is operational for two weeks, seems to work :+1: 
Would this warrant a new release? :angel:

spuetz:
Sure, end of the week I'll make a new version. Cheers.

PR #263, opened 2021-03-04
fix controller termination

dorezyuk:
With the #262 merge, we have introduced a regression in the termination criteria of our controllers: 

If the user doesn't specify max_rettries_ or patience_ we would not terminate at all. I've made the behavior congruent with the behavior of abstract_planner_execution - terminating the loop if both parameters are disabled. 

Keep in mind that the #261 PR will provide some testing the affected class (but I would need to adjust the tests to the changed behavior).

Best,
Dima


spuetz:
Is this the adjustment you meant? :) https://github.com/magazino/move_base_flex/runs/2113351121 We should set a retry value here, or?

Issue #260, opened 2021-02-19
Transitioning goal lost cause move_base_flex die


echo4echo:
I used the move_base_flex in my project and found a issue in my test. When the robot was sitting the "Black zone" in the global map (That would happen when robot localization is not excatly correct or manually mis-operating), after exceeding the patience time, the move_base_flex would die.
![Screenshot from 2021-02-18 16-20-54](https://user-images.githubusercontent.com/39905476/108447424-ccb6e380-729a-11eb-80c5-4797c1a51da1.png)

dorezyuk:
Thanks for submitting. Could you upload the coredump from the node?

corot:
@echo4echo, can u retry with current master?  I think this issue should be fixed now

echo4echo:
@corot 
Hey, Corot, sorry for the late reply. Currently our robot is under maintenance. I will pull down and rerun again when I get the time slot to use the robot.

corot:
Hopefully solved with #267; @echo4echo, please reopen if you still find the issue with current master

PR #267, opened 2021-04-20
Run replanning in its own thread (fixes #223)

### LINKED ISSUE #223: Canceling move_base action takes very long when doing continuous replanning
Looks like it takes approximately replanning period! All happens in this line:

```
  if(!action_client_get_path_.getState().isDone())
```
Most probably is caused by the calls to replanning_rate_.sleep(), so I suppose we must move the replanning to a separated thread.

### LINKED ISSUE #259: Actionlib bug when planner_frequency is anything other than 0
**ROS:** Noetic
**Package:** ros-noetic-move-base-flex 0.3.4-1focal.20210112.101859

**planners.yaml:**
```
planners:
  - name: 'planner'
    type: 'global_planner/GlobalPlanner'
```

I get the following console output when setting the global planner_frequency to anything other than 0hz:

```
BUG: Got a transition to CommState [ACTIVE] when in SimpleGoalState [DONE]
BUG: Got a second transition to DONE
```

The global planner only ends up publishing a total of 2 times /to `move_base_flex/planner/plan` and on the second time it outputs the above messages to the console.

This bug only happens only once per navigation goal. The robot does end up reaching its target however, and I get the `GOAL Reached!` message successfully.

corot:
As I mentioned in the issue, we need to run the replanning in its own thread

Interestingly, it also seems to fix #259 and #260 (both the same issue, I think)

@echo4echo, @r0gi, would be really great if you can give  a try to this branch and see if it fixes the issue you reported;  thanks in advance

### DIFF
--- a/mbf_abstract_nav/src/move_base_action.cpp
+++ b/mbf_abstract_nav/src/move_base_action.cpp
@@ -38,6 +38,8 @@
  *
  */
 
+#include <limits>
+
 #include <mbf_utility/navigation_utility.h>
 
 #include "mbf_abstract_nav/MoveBaseFlexConfig.h"
@@ -66,6 +68,8 @@ MoveBaseAction::MoveBaseAction(const std::string& name, const mbf_utility::Robot
 
 MoveBaseAction::~MoveBaseAction()
 {
+  action_state_ = NONE;
+  replanning_thread_.join();
 }
 
 void MoveBaseAction::reconfigure(
@@ -509,23 +513,26 @@ void MoveBaseAction::actionRecoveryDone(
   }
 }
 
+bool MoveBaseAction::replanningActive() const
+{
+  // replan only while following a path and if replanning is enabled (can be disabled by dynamic reconfigure)
+  return !replanning_period_.isZero() && action_state_ == EXE_PATH && dist_to_goal_ > 0.1;
+}
+
 void MoveBaseAction::replanningThread()
 {
-  ros::Duration update_period(0.001);
+  ros::Duration update_period(0.005);
   ros::Time last_replan_time(0.0);
 
   while (ros::ok())
   {
-    // replan only while following a path and if replanning is enabled (can be disabled by dynamic reconfigure)
-    bool keep_replanning = !replanning_period_.isZero() && action_state_ == EXE_PATH && dist_to_goal_ > 0.1;
-
     if (!action_client_get_path_.getState().isDone())
     {
       if (action_client_get_path_.waitForResult(update_period))
       {
         actionlib::SimpleClientGoalState state = action_client_get_path_.getState();
         mbf_msgs::GetPathResultConstPtr result = action_client_get_path_.getResult();
-        if (state == actionlib::SimpleClientGoalState::SUCCEEDED && keep_replanning)
+        if (state == actionlib::SimpleClientGoalState::SUCCEEDED && replanningActive())
         {
           ROS_DEBUG_STREAM_NAMED("move_base", "Replanning succeeded; sending a goal to \"exe_path\" with the new plan");
           exe_path_goal_.path = result->path;
@@ -540,20 +547,13 @@ void MoveBaseAction::replanningThread()
                                  "Replanning failed with error code " << result->outcome << ": " << result->message);
         }
       }
-      else
-      {
-        // keep waiting for planning to complete (we already waited update_period in waitForResult)
-        continue;
-      }
+      // else keep waiting for planning to complete (we already waited update_period in waitForResult)
     }
-
-    if (!keep_replanning)
+    else if (!replanningActive())
     {
       update_period.sleep();
-      continue;
     }
-
-    if (ros::Time::now() - last_replan_time >= replanning_period_)
+    else if (ros::Time::now() - last_replan_time >= replanning_period_)
     {
       ROS_DEBUG_STREAM_NAMED("move_base", "Next replanning cycle, using the \"get_path\" action!");
       action_client_get_path_.sendGoal(get_path_goal_);
--- a/mbf_abstract_nav/src/move_base_action.cpp
+++ b/mbf_abstract_nav/src/move_base_action.cpp
@@ -45,22 +45,22 @@
 
 namespace mbf_abstract_nav
 {
-
-MoveBaseAction::MoveBaseAction(const std::string &name,
-                               const mbf_utility::RobotInformation &robot_info,
-                               const std::vector<std::string> &behaviors)
-  :  name_(name), robot_info_(robot_info), private_nh_("~"),
-     action_client_exe_path_(private_nh_, "exe_path"),
-     action_client_get_path_(private_nh_, "get_path"),
-     action_client_recovery_(private_nh_, "recovery"),
-     oscillation_timeout_(0),
-     oscillation_distance_(0),
-     recovery_enabled_(true),
-     behaviors_(behaviors),
-     action_state_(NONE),
-     recovery_trigger_(NONE),
-     replanning_(false),
-     replanning_rate_(1.0)
+MoveBaseAction::MoveBaseAction(const std::string& name, const mbf_utility::RobotInformation& robot_info,
+                               const std::vector<std::string>& behaviors)
+  : name_(name)
+  , robot_info_(robot_info)
+  , private_nh_("~")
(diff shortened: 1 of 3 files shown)


mbf_abstract_nav/src/move_base_action.cpp
```diff
@@ -45,22 +45,22 @@
...
+  , action_client_exe_path_(private_nh_, "exe_path")
+  , action_client_get_path_(private_nh_, "get_path")
+  , action_client_recovery_(private_nh_, "recovery")
+  , oscillation_timeout_(0)
+  , oscillation_distance_(0)
+  , recovery_enabled_(true)
+  , behaviors_(behaviors)
+  , action_state_(NONE)
+  , recovery_trigger_(NONE)
+  , dist_to_goal_(std::numeric_limits<double>::infinity())
```
corot, line 64: Only this is new; the rest is automatic clang reformatting of the modified code

spuetz (approved):
lgtm

mbf_abstract_nav/src/move_base_action.cpp
```diff
@@ -121,6 +102,8 @@ void MoveBaseAction::cancel()
 
 void MoveBaseAction::start(GoalHandle &goal_handle)
 {
+  dist_to_goal_ = std::numeric_limits<double>::infinity();
```
dorezyuk, line 109: add include limits

mbf_abstract_nav/src/move_base_action.cpp
```diff
@@ -45,22 +45,22 @@
...
+  , oscillation_timeout_(0)
+  , oscillation_distance_(0)
+  , recovery_enabled_(true)
+  , behaviors_(behaviors)
+  , action_state_(NONE)
+  , recovery_trigger_(NONE)
+  , dist_to_goal_(std::numeric_limits<double>::infinity())
+  , replanning_thread_(boost::bind(&MoveBaseAction::replanningThread, this))
 {
 }
```
dorezyuk, line 68: we should join the replanning_thread in the destructor

mbf_abstract_nav/src/move_base_action.cpp
```diff
@@ -544,36 +509,57 @@ void MoveBaseAction::actionRecoveryDone(
...
-  }
+    // replan only while following a path and if replanning is enabled (can be disabled by dynamic reconfigure)
+    bool keep_replanning = !replanning_period_.isZero() && action_state_ == EXE_PATH && dist_to_goal_ > 0.1;
 
-  replanning_mtx_.lock();
-  replanning_rate_.sleep();
-  replanning_mtx_.unlock();
+    if (!action_client_get_path_.getState().isDone())
+    {
+      if (action_client_get_path_.waitForResult(update_period))
```
dorezyuk, line 531: why do we wait here only so little? did you had issues with shutting down the thread correctly?

mbf_abstract_nav/src/move_base_action.cpp
```diff
@@ -544,36 +509,57 @@ void MoveBaseAction::actionRecoveryDone(
...
+      {
+        actionlib::SimpleClientGoalState state = action_client_get_path_.getState();
+        mbf_msgs::GetPathResultConstPtr result = action_client_get_path_.getResult();
+        if (state == actionlib::SimpleClientGoalState::SUCCEEDED && keep_replanning)
+        {
+          ROS_DEBUG_STREAM_NAMED("move_base", "Replanning succeeded; sending a goal to \"exe_path\" with the new plan");
```
dorezyuk, line 537: I think we have here a potential race condition:

When we reach this point keep_replanning might be outdated (it seem to me that this is even likely to happen) and we might then either send a goal to the controller while recovering or when the user/our logic already canceled the current execution

mbf_abstract_nav/src/move_base_action.cpp
```diff
@@ -544,36 +509,57 @@ void MoveBaseAction::actionRecoveryDone(
...
+    {
+      update_period.sleep();
+      continue;
+    }
 
-  ROS_DEBUG_STREAM_NAMED("move_base", "Next replanning cycle, using the \"get_path\" action!");
-  action_client_get_path_.sendGoal(
-      get_path_goal_,
-      boost::bind(&MoveBaseAction::actionGetPathReplanningDone, this, _1, _2)); // replanning
+    if (ros::Time::now() - last_replan_time >= replanning_period_)
```
dorezyuk: does it make sense to check the state of action_client_get_path_ - I'm thinking that if the user picks a replanning period shorted than the planners response time we will always preempt the planner

mbf_abstract_nav/src/move_base_action.cpp
```diff
@@ -544,36 +509,57 @@ void MoveBaseAction::actionRecoveryDone(
...
+        // keep waiting for planning to complete (we already waited update_period in waitForResult)
+        continue;
+      }
+    }
 
-  if (!replanning_ || action_state_ != EXE_PATH)
-    return; // another chance to stop replannings after waiting for replanning period
+    if (!keep_replanning)
+    {
+      update_period.sleep();
```
dorezyuk, line 554: what about using a condition var here with a bigger timeout? we can trigger the condition var every time we set the state to exe_path or change the replanning period

mbf_abstract_nav/src/move_base_action.cpp
dorezyuk: just personal preference but it seems better readable if we make the ifs else ifs:

```
if(!action_client_get_path_.getState().isDone()){...}
else if(!keep_replanning) {...}
else if(ros::Time::now() - last_replan_time >= replanning_period_) {...}
```
in this manner we would also resolve the issue with preempting the planer

mbf_abstract_nav/src/move_base_action.cpp
corot: no particular reason,,, 1ms sounds reasonable to react fast to changes on state
will inc x5

mbf_abstract_nav/src/move_base_action.cpp
corot: mmm...  yes,,, maybe that's why I set the loop at 1ms.   Probably better now

spuetz:
You are right with the preempting. I like your proposals. This could be done in another PR, or? What do you think @dorezyuk @corot  ?

mbf_abstract_nav/src/move_base_action.cpp
corot: > does it make sense to check the state of action_client_get_path_ - I'm thinking that if the user picks a replanning period shorted than the planners response time we will always preempt the planner

nope,,,  we will continue [here](https://github.com/magazino/move_base_flex/pull/267/files#diff-6b8a12259c3d11b5f68c080c8a6d7a690aecea38015c66366f1d559b16bd4acbR546) and so not send a new goal

dorezyuk: ah yea you're right - my bad

corot:
> You are right with the preempting. I like your proposals. This could be done in another PR, or? What do you think @dorezyuk @corot ?

I'm done with the requested changes;  just need to test
I'll commit now, before testing, so u can judge

mbf_abstract_nav/src/move_base_action.cpp
```diff
@@ -544,36 +511,53 @@ void MoveBaseAction::actionRecoveryDone(
   }
 }
 
-void MoveBaseAction::actionGetPathReplanningDone(
-    const actionlib::SimpleClientGoalState &state,
-    const mbf_msgs::GetPathResultConstPtr &result)
+bool MoveBaseAction::replanningActive()
```
dorezyuk: can you make it const?

mbf_abstract_nav/src/move_base_action.cpp
dorezyuk: @corot can you fix this - then the pr is fine :)

dorezyuk:
> > You are right with the preempting. I like your proposals. This could be done in another PR, or? What do you think @dorezyuk @corot ?
> 
> I'm done with the requested changes; just need to test
> I'll commit now, before testing, so u can judge

It looks good now - the only thing I could wish for are some tests for this class;)

dorezyuk (approved):

PR #270, opened 2021-06-11
Fix #269: do not reset cancel_, as that will invalidate ongoing canceling

### LINKED ISSUE #269: Cancel exe_path action is ignored if sent < 100 ms after the goal
To reproduce, set shutdown_costmaps to true, as this introduce a delay between the exe_path goal reception and the execution.

If you send a cancel soon (<100ms) after the goal, we will set the cancel_ flag to true [here](https://github.com/magazino/move_base_flex/blob/master/mbf_abstract_nav/src/abstract_controller_execution.cpp#L263), but then the execution start reset it to false [here](https://github.com/magazino/move_base_flex/blob/master/mbf_abstract_nav/src/abstract_execution_base.cpp#L67).

I tried to remove these 3 lines under "reset the member vars" and now cancel works regardless of how fast u send it after the goal. I think these lines are not needed, as we create a new execution object each time. @spuetz, can u confirm?

corot:
As the controller-execution object will be created for a start event, we can safely remove these lines

### DIFF
--- a/mbf_abstract_nav/src/abstract_execution_base.cpp
+++ b/mbf_abstract_nav/src/abstract_execution_base.cpp
@@ -63,11 +63,6 @@ bool AbstractExecutionBase::start()
     thread_.join();
   }
 
-  // reset the member vars
-  cancel_ = false;
-  outcome_ = 255;
-  message_.clear();
-
   thread_ = boost::thread(&AbstractExecutionBase::run, this);
   return true;
 }
@@ -78,12 +73,13 @@ void AbstractExecutionBase::stop()
   thread_.interrupt();
 }
 
-void AbstractExecutionBase::join(){
+void AbstractExecutionBase::join()
+{
   if (thread_.joinable())
     thread_.join();
 }
 
-boost::cv_status AbstractExecutionBase::waitForStateUpdate(boost::chrono::microseconds const &duration)
+boost::cv_status AbstractExecutionBase::waitForStateUpdate(boost::chrono::microseconds const& duration)
 {
   boost::mutex mutex;
   boost::unique_lock<boost::mutex> lock(mutex);


mbf_abstract_nav/src/abstract_execution_base.cpp
```diff
@@ -78,12 +73,13 @@ void AbstractExecutionBase::stop()
   thread_.interrupt();
 }
 
-void AbstractExecutionBase::join(){
+void AbstractExecutionBase::join()
```
corot, line 76: otherwise clang complains,,, which is weird, cause this is well away from the changes  :thinking:

spuetz (approved):

mbf_abstract_nav/src/abstract_execution_base.cpp
```diff
@@ -63,11 +63,6 @@ bool AbstractExecutionBase::start()
     thread_.join();
   }
 
-  // reset the member vars
-  cancel_ = false;
```
dorezyuk, line 67: We never reuse right now the execution class, so this is fine.

However, I think we should make the class "reusable" (or prevent explicitly the users from reusing an instance). This might be a cool application for our preRun and postRun methods:

```
virtual void preRun() {
// clear the old state:
outcome_ = 255;
message_.clear();
}
```
and
```
virtual void postRun(){
// reset the cancel flag
cancel_ = false
}
```
What do you guys think?

corot: > We never reuse right now the execution class, so this is fine.
> 
> However, I think we should make the class "reusable" (or prevent explicitly the users from reusing an instance). This might be a cool application for our preRun and postRun methods:

The idea of pre/postRun was other, namely calling map-specific code before/after the execution. Clearing cancel flag is a bit out of place, imho. In any case, I propose to wait until we make the executions reusable to made that change.

if you are both ok with this PR, i'll merge it as it is

dorezyuk (approved):

PR #272, opened 2021-07-07
return empty footprint if partly outside of the map

dorezyuk:
Addresses the issues of #271

### DIFF
--- a/mbf_costmap_nav/include/mbf_costmap_nav/footprint_helper.h
+++ b/mbf_costmap_nav/include/mbf_costmap_nav/footprint_helper.h
@@ -62,11 +62,9 @@ public:
    * @param  fill If true: returns all cells in the footprint of the robot. If false: returns only the cells that make up the outline of the footprint.
    * @return The cells that make up either the outline or entire footprint of the robot depending on fill
    */
-  static std::vector<Cell> getFootprintCells(
-      double x, double y, double theta,
-      std::vector<geometry_msgs::Point> footprint_spec,
-      const costmap_2d::Costmap2D&,
-      bool fill);
+  static std::vector<Cell> getFootprintCells(double x, double y, double theta,
+                                             const std::vector<geometry_msgs::Point>& footprint_spec,
+                                             const costmap_2d::Costmap2D&, bool fill);
 
   /**
    * @brief  Use Bresenham's algorithm to trace a line between two points in a grid
--- a/mbf_costmap_nav/src/mbf_costmap_nav/footprint_helper.cpp
+++ b/mbf_costmap_nav/src/mbf_costmap_nav/footprint_helper.cpp
@@ -167,13 +167,21 @@ void FootprintHelper::getFillCells(std::vector<Cell>& footprint){
 }
 
 /**
+ * @brief Helper function which clears and returns the input.
+ */
+static const std::vector<Cell>& clearAndReturn(std::vector<Cell>& _cells)
+{
+  _cells.clear();
+  return _cells;
+}
+
+/**
  * get the cells of a footprint at a given position
  */
-std::vector<Cell> FootprintHelper::getFootprintCells(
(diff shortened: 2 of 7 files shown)


mbf_utility/src/navigation_utility.cpp
```diff
@@ -69,12 +71,31 @@ bool getRobotPose(const TF &tf,
   return success;
 }
 
+/**
+ * @brief Returns true, ifthe given Quaternion is normalized.
+ *
+ * @param _q The quaternion to check.
+ * @param _epsilon The epsilon (squared distance to 1).
+ */
+static bool isNormalized(const geometry_msgs::Quaternion& _q, double _epsilon)
```
dorezyuk, line 80: without this we would return also empty if the quaternion would be invalid (then tf2::getYaw would return nan)

mbf_costmap_nav/include/mbf_costmap_nav/footprint_helper.h
```diff
@@ -62,11 +62,9 @@ class FootprintHelper
    * @param  fill If true: returns all cells in the footprint of the robot. If false: returns only the cells that make up the outline of the footprint.
    * @return The cells that make up either the outline or entire footprint of the robot depending on fill
    */
-  static std::vector<Cell> getFootprintCells(
-      double x, double y, double theta,
-      std::vector<geometry_msgs::Point> footprint_spec,
```
dorezyuk, line 67: made it passed by const ref

corot (changes requested):
yup, works as expected
just need to change a bit the comments, also in the comment on CheckPose/CheckPath services ( I did )

mbf_utility/src/navigation_utility.cpp
dorezyuk: indirectly - if the quaternion is ill formed we would silently return "out of costmap bounds" instead of telling what's wrong

mbf_costmap_nav/src/mbf_costmap_nav/footprint_helper.cpp
```diff
@@ -200,14 +208,14 @@ std::vector<Cell> FootprintHelper::getFootprintCells(
     new_x = x + (footprint_spec[i].x * cos_th - footprint_spec[i].y * sin_th);
     new_y = y + (footprint_spec[i].x * sin_th + footprint_spec[i].y * cos_th);
     if(!costmap.worldToMap(new_x, new_y, x0, y0)) {
```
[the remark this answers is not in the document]
dorezyuk: The re-formatting only applies to the lines changed in a pr

mbf_msgs/srv/CheckPath.srv
```diff
@@ -12,11 +12,11 @@ uint8                      skip_poses        # skip this number of poses between
 bool                       path_cells_only   # check only cells directly traversed by the path, ignoring robot footprint
                                              # (if true, safety_dist is ignored)
 ---
-uint8                      FREE      =  0    # robot is completely in traversable space
```
dorezyuk, line 15: would this change the md5 sum of the message?

mbf_utility/src/navigation_utility.cpp
dorezyuk: Would you rather move this change into a dedicated PR?

corot: should be, but I'm ok with merging it here (I'm not a purist)

mbf_msgs/srv/CheckPath.srv
corot: looks like not, as comments are removed before serializing
http://wiki.ros.org/ROS/Technical%20Overview#Message_serialization_and_msg_MD5_sums

corot (approved):

spuetz (approved):

PR #273, opened 2021-07-08
Lock less aggressively slot_map_mtx_, so sending a new goal when canceling doesn't freeze MBF

corot:
We are locking the access to concurrency_slots_ map when doing any operation, as start and cancel. That's not needed. Interators to map [don't get invalidated by adding new entries to the map](https://www.cplusplus.com/reference/map/map/insert/), and we are not removing concurrency slots other than in shutdown.

With the current master, if we receive a new goal while performing a lengthy cancel (e.g. because we want to handle a smooth stop, as requested on [this PR](https://github.com/magazino/move_base_flex/pull/172)), MBF will freeze, as we lock twice slot_map_mtx_, [here](https://github.com/magazino/move_base_flex/pull/273/files#diff-9dd967a72350032851289ad39d52abbf057bc9ed6c90f61b0a06c5bd673c1fbaL140) and [here](https://github.com/magazino/move_base_flex/pull/273/files#diff-9dd967a72350032851289ad39d52abbf057bc9ed6c90f61b0a06c5bd673c1fbaL178).

With this PR, we accept the new goal, wait for the previous cancel to complete, and then execute the new goal.

### DIFF
--- a/mbf_abstract_nav/include/mbf_abstract_nav/abstract_action_base.hpp
+++ b/mbf_abstract_nav/include/mbf_abstract_nav/abstract_action_base.hpp
@@ -138,9 +138,7 @@ public:
     }
     else
     {
-      slot_map_mtx_.lock();
-      typename ConcurrencyMap::iterator slot_it = concurrency_slots_.find(slot);
-      slot_map_mtx_.unlock();
+      typename ConcurrencyMap::iterator slot_it = findSlot(slot);
       if (slot_it != concurrency_slots_.end() && slot_it->second.in_use)
       {
         // if there is already a plugin running on the same slot, cancel it
@@ -179,9 +177,7 @@ public:
   {
     uint8_t slot = goal_handle.getGoal()->concurrency_slot;
 
-    slot_map_mtx_.lock();
-    typename ConcurrencyMap::iterator slot_it = concurrency_slots_.find(slot);
-    slot_map_mtx_.unlock();
+    typename ConcurrencyMap::iterator slot_it = findSlot(slot);
     if (slot_it != concurrency_slots_.end() && slot_it->second.execution)
     {
       slot_it->second.execution->cancel();
@@ -210,7 +206,8 @@ public:
     typename ConcurrencyMap::iterator iter;
     for (iter = concurrency_slots_.begin(); iter != concurrency_slots_.end(); ++iter)
     {
-      iter->second.execution->reconfigure(config);
+      if (iter->second.execution)
+        iter->second.execution->reconfigure(config);
     }
   }
 
@@ -222,7 +219,8 @@ public:
     typename ConcurrencyMap::iterator iter;
     for (iter = concurrency_slots_.begin(); iter != concurrency_slots_.end(); ++iter)
     {
-      iter->second.execution->cancel();
+      if (iter->second.execution)
+        iter->second.execution->cancel();
     }
     threads_.join_all();
   }
@@ -236,6 +234,11 @@ protected:
 
   boost::mutex slot_map_mtx_;
 
+  typename ConcurrencyMap::iterator findSlot(uint8_t slot)
+  {
+    boost::lock_guard<boost::mutex> guard(slot_map_mtx_);
+    return concurrency_slots_.find(slot);
+  }
 };
 
 }
--- a/mbf_abstract_nav/include/mbf_abstract_nav/abstract_action_base.hpp
+++ b/mbf_abstract_nav/include/mbf_abstract_nav/abstract_action_base.hpp
@@ -110,6 +110,7 @@ public:
     // cleanup threads used on executions
     // note: cannot call cancelAll, since our mutex is not recursive
     boost::lock_guard<boost::mutex> guard(slot_map_mtx_);
+
     typename ConcurrencyMap::iterator slot_it = concurrency_slots_.begin();
     for (; slot_it != concurrency_slots_.end(); ++slot_it)
     {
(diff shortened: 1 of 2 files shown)


mbf_abstract_nav/include/mbf_abstract_nav/abstract_action_base.hpp
```diff
@@ -131,28 +132,31 @@ class AbstractActionBase
...
       typename ConcurrencyMap::iterator slot_it = concurrency_slots_.find(slot);
-      if (slot_it != concurrency_slots_.end() && slot_it->second.in_use) {
+      slot_map_mtx_.unlock();
+      if (slot_it != concurrency_slots_.end() && slot_it->second.in_use)
+      {
         // if there is already a plugin running on the same slot, cancel it
         slot_it->second.execution->cancel();
 
-        if (slot_it->second.thread_ptr->joinable()) {
+        if (slot_it->second.thread_ptr->joinable())
```
dorezyuk, line 147: not sure why you formatted this. did you run `git clang-format master`?

dorezyuk:
I like the idea, but the mutex also protected us from concurrent access to the members of the iterator. 

From the quick look we might for example run into problems when calling start and cancel "concurrently" (assuming we first call start).
```
// inside AbtractActionBase::start
// if slot_it is newly created the execution is nullptr
      slot_it->second.execution = execution_ptr;
```
```
// inside AbtractActionBase::cancel
    concurrency_slots_[slot].execution->cancel();
```
would be a nullptr deref.

If its not a nullptr, we still might call cancel on a execution, which hasn't started and would then miss the cancel

mbf_abstract_nav/include/mbf_abstract_nav/abstract_action_base.hpp
corot: The idea was to format little by little with the PRs, right?
I can undo, anyway

dorezyuk: sure, but not at random :grin:

corot:
> If its not a nullptr, we still might call cancel on a execution, which hasn't started and would then miss the cancel

check added.  we can take care of the possible corner cases; but getting frozen as now is not an option!   :skull:

mbf_abstract_nav/include/mbf_abstract_nav/abstract_action_base.hpp
```diff
@@ -203,7 +208,7 @@ class AbstractActionBase
     boost::lock_guard<boost::mutex> guard(slot_map_mtx_);
 
     typename ConcurrencyMap::iterator iter;
-    for(iter = concurrency_slots_.begin(); iter != concurrency_slots_.end(); ++iter)
+    for (iter = concurrency_slots_.begin(); iter != concurrency_slots_.end(); ++iter)
     {
       iter->second.execution->reconfigure(config);
```
dorezyuk: this might be a null ptr access

mbf_abstract_nav/include/mbf_abstract_nav/abstract_action_base.hpp
```diff
@@ -213,8 +218,9 @@ class AbstractActionBase
   {
     ROS_INFO_STREAM_NAMED(name_, "Cancel all goals for \"" << name_ << "\".");
     boost::lock_guard<boost::mutex> guard(slot_map_mtx_);
+
     typename ConcurrencyMap::iterator iter;
-    for(iter = concurrency_slots_.begin(); iter != concurrency_slots_.end(); ++iter)
+    for (iter = concurrency_slots_.begin(); iter != concurrency_slots_.end(); ++iter)
     {
       iter->second.execution->cancel();
```
dorezyuk: this might also be a nullptr

mbf_abstract_nav/include/mbf_abstract_nav/abstract_action_base.hpp
```diff
@@ -175,11 +179,12 @@ class AbstractActionBase
...
 
-    boost::lock_guard<boost::mutex> guard(slot_map_mtx_);
+    slot_map_mtx_.lock();
     typename ConcurrencyMap::iterator slot_it = concurrency_slots_.find(slot);
-    if (slot_it != concurrency_slots_.end())
+    slot_map_mtx_.unlock();
+    if (slot_it != concurrency_slots_.end() && slot_it->second.execution)
     {
-      concurrency_slots_[slot].execution->cancel();
+      slot_it->second.execution->cancel();
```
dorezyuk, line 183: while we have ruled out that this will be a nullptr, it seems to me that we still might face a race condition  the cancel might be called before the start - missing the cancel request

mbf_abstract_nav/include/mbf_abstract_nav/abstract_action_base.hpp
```diff
@@ -175,11 +179,12 @@ class AbstractActionBase
   {
     uint8_t slot = goal_handle.getGoal()->concurrency_slot;
 
-    boost::lock_guard<boost::mutex> guard(slot_map_mtx_);
+    slot_map_mtx_.lock();
     typename ConcurrencyMap::iterator slot_it = concurrency_slots_.find(slot);
```
dorezyuk: just as an idea: would it make sense to write a function which looks like
```
typename ConcurrencyMap::iterator AbstractActionBase::findSlot(uint_8 _key) {
 boost::scoped_lock<...> lock;
return concurrency_slots.find(_key)
}

```

dorezyuk:
@corot  Do you mean with "freeze" as in deadlock (the node then just sits there and never recovers)? If yes, can you explain how this is happening?

mbf_abstract_nav/include/mbf_abstract_nav/abstract_action_base.hpp
corot: yes, I like it

mbf_abstract_nav/include/mbf_abstract_nav/abstract_action_base.hpp
corot: We need to define what's the expected results if a new goal comes right after the cancel. In the particular case you mention, I think the new goal superseeds the cancel and we proceed to execute it

corot:
> @corot Do you mean with "freeze" as in deadlock (the node then just sits there and never recovers)? If yes, can you explain how this is happening?

Yep, the node freezes and don't even react to rosnode kill signal. To be honest, I cannot explain how this happens, because even when we return the cancel, we don't recover (I would expect that the new goal starts then, as the lock is released,,, but somehow this doesn't happen!   we never release the lock)

Reproducing is easy, though. Implement the cancel method of your controller, e.g.

```
  bool cancel() {
    for (int i = 0; i < 50; ++i) {
      ROS_INFO("cancel %d...",i+1/5);
      ros::spinOnce();
      ros::Duration(0.1).sleep();

    }
    ROS_INFO("cancel DONE!");
    return true;
  };
```
The spinOnce is important, because otherwise MBF won't call start and cancel simultaneously. This sounds like solving the problem, but then your controller won't get ANY info, e.g. velocity feedback.

Then u send any path to exe_path, cancel it and send it again within 5 seconds

corot:
After reconsidering this PR, I have another proposal for handling smooth cancel, without the need of blocking the cancel callback #274. Early feedback more than welcome!! :pray:

corot:
Closed in favor of #274

PR #274, opened 2021-07-12
Allow the controller to handle cancel if properly implemented

corot:
Possibly a much safer solution for #273. This allows smooth cancels with decreasing velocities without risking the deadlocks reported in #273. The controller becomes responsible of slowing down by returning decreasing speeds on computeVelocityCommands and SUCCESS outcome. When he considers that the robot can stop, returns zero velocity and CANCELED outcome.

Trivial example implementation for TEB:

image

If the controller doesn't implement the controller plugin API cancel method, or returns false, we use the normal stop method with an abrupt zero-velocity.

:warning::warning::warning:
One corner case is when we cancel the controller but it reaches the goal while slowing down. In this case, we will report success, which is a sensible result because we accomplish the requested goal despite the cancel, but one could argue that we should return preempted.
:warning::warning::warning:

Example:

1. send goal to exe_path
2. cancel it, slowing down
3. send the same goal before it stops  --> we preempt the canceling and start executing the new goal

https://user-images.githubusercontent.com/322610/125382572-d91a1500-e3d0-11eb-92fc-cc8776a2a791.mp4

### DIFF
--- a/mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
+++ b/mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
@@ -119,10 +119,12 @@ namespace mbf_abstract_nav
       double action_angle_tolerance = 3.1415);
 
     /**
-     * @brief Cancel the controller execution.
-     * This calls the cancel method of the controller plugin, sets the cancel_ flag to true,
-     * and waits for the control loop to stop. Normally called upon aborting the navigation.
-     * @return true, if the control loop stops within a cycle time.
+     * @brief Cancel the controller execution. Normally called upon aborting the navigation.
+     * This calls the cancel method of the controller plugin. If the plugins returns true, it becomes
+     * responsible of stopping, and we will keep requesting velocity commands until it returns CANCELED.
+     * If it returns false (meaning cancel is not implemented, or that the controller defers handling it),
+     * MBF will set the cancel_ flag to true, and wait for the control loop to stop.
+     * @return true, if the controller handles the stooping, or the control loop stops within a cycle time.
      */
     virtual bool cancel();
 
--- a/mbf_abstract_nav/src/abstract_controller_execution.cpp
+++ b/mbf_abstract_nav/src/abstract_controller_execution.cpp
@@ -254,18 +254,24 @@ bool AbstractControllerExecution::reachedGoalCheck()
 
 bool AbstractControllerExecution::cancel()
 {
-  // request the controller to cancel; it returns false if cancel is not implemented or rejected by the plugin
-  if (!controller_->cancel())
+  // Request the controller to cancel; it will return true if it takes care of stopping, returning CANCELED on
+  // computeVelocityCmd when done. This allows for smooth, controlled stops.
+  // If false (meaning cancel is not implemented, or that the controller defers handling it) MBF will take care.
+  if (controller_->cancel())
   {
-    ROS_WARN_STREAM("Cancel controlling failed. Wait until the current control cycle finished!");
+    ROS_INFO("Controller will take care of stopping");
   }
-  // then wait for the control cycle to stop (should happen immediately if the controller cancel returned true)
-  cancel_ = true;
-  if (waitForStateUpdate(boost::chrono::milliseconds(500)) == boost::cv_status::timeout)
+  else
   {
-    // this situation should never happen; if it does, the action server will be unready for goals immediately sent
-    ROS_WARN_STREAM("Timeout while waiting for control cycle to stop; immediately sent goals can get stuck");
-    return false;
+    ROS_WARN("Controller defers handling cancel; force it and wait until the current control cycle finished");
+    cancel_ = true;
+    // wait for the control cycle to stop
+    if (waitForStateUpdate(boost::chrono::milliseconds(500)) == boost::cv_status::timeout)
+    {
+      // this situation should never happen; if it does, the action server will be unready for goals immediately sent
+      ROS_WARN_STREAM("Timeout while waiting for control cycle to stop; immediately sent goals can get stuck");
+      return false;
+    }
   }
   return true;
 }
@@ -381,6 +387,12 @@ void AbstractControllerExecution::run()
           vel_pub_.publish(cmd_vel_stamped.twist);
           last_valid_cmd_time_ = ros::Time::now();
           retries = 0;
+        }
+        else if (outcome_ == mbf_msgs::ExePathResult::CANCELED)
+        {
+          ROS_INFO_STREAM("Controller-handled cancel completed");
+          cancel_ = true;
+          continue;
         }
         else
         {


dorezyuk:
@corot I had a look at the bug and it looks not good.  I've dead locked the node with your code and generated a coredump. Interestingly in the coredump, I didn't find the thread execution the cancel method anymore - which explains why the cancel method never returns...

Since its related to the spinning (I'm not sure if spinning within the plugin is actually a good idea but this is not the point), I've created a dedicated callback queue just for our servers - and this seems to solve the problem.

Could you check out https://github.com/magazino/move_base_flex/tree/dima/callback_queue (without the here mentioned changed to TEB) to see if it works for you?

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -254,18 +254,24 @@ bool AbstractControllerExecution::reachedGoalCheck()
...
-  // then wait for the control cycle to stop (should happen immediately if the controller cancel returned true)
-  cancel_ = true;
-  if (waitForStateUpdate(boost::chrono::milliseconds(500)) == boost::cv_status::timeout)
+  else
   {
-    // this situation should never happen; if it does, the action server will be unready for goals immediately sent
-    ROS_WARN_STREAM("Timeout while waiting for control cycle to stop; immediately sent goals can get stuck");
-    return false;
+    ROS_WARN("Controller defers handling cancel; force it and wait until the current control cycle finished");
+    cancel_ = true;
```
dorezyuk, line 267: should we set it also for the other case?

corot: in the other case we set it [here](https://github.com/magazino/move_base_flex/pull/274/files#diff-bbabf6c9cc39ada410c0f58e40f9cf9c80a104f403ee9196f6a1519fe972a2dfR394), once we are done with the cancel

corot:
> @corot I had a look at the bug and it looks not good. I've dead locked the node with your code and generated a coredump. Interestingly in the coredump, I didn't find the thread execution the cancel method anymore - which explains why the cancel method never returns...

you mean, u had a deadlock with this PR?   And which controller?  default TEB or my example for slowing down?

corot:
> Interestingly in the coredump, I didn't find the thread execution the cancel method anymore - which explains why the cancel method never returns...

wooooo,,,,  I knew something superfishy was happening,,, I though because of stopping the execution thread, but not
any idea why a thread can disappear like that?

dorezyuk:
> you mean, u had a deadlock with this PR? And which controller? default TEB or my example for slowing down?

Your TEB example from #273

mbf_abstract_nav/src/abstract_controller_execution.cpp
dorezyuk: I was thinking just to tolerate if the user implements the interface wrong (say, returning SUCCESS instead of CANCEL)

corot:
> Your TEB example from #273

Ah,,,  yes, yes.  I thought u meant this PR  :relieved: 

It's still interesting to investigate that strange problem, but this PR is a much cleaner solution to smooth cancels, imho

corot:
> Could you check out https://github.com/magazino/move_base_flex/tree/dima/callback_queue (without the here mentioned changed to TEB) to see if it works for you?

But wait,,, this can be much simpler, right?  Not enough to [run an async spinner instead of spin](https://github.com/magazino/move_base_flex/commit/89e87d2a02ab339eb81c94afb21e49836abee702)?

In any case, yes, sounds reasonable to run this multi-threaded. Did you try to reproduce the freezing with this change, but removing the `ros::spinOnce();`?

dorezyuk:
> But wait,,, this can be much simpler, right? Not enough to run an async spinner instead of spin?

Also a good idea - to me both seems fine.

> Did you try to reproduce the freezing with this change, but removing the ros::spinOnce();?

Not sure about which spinOnce we are talking now :confused: About the one in cancel or in the main? It worked for me *with* the spinOnce in the cancel

corot:
> Not sure about which spinOnce we are talking now  About the one in cancel or in the main? It worked for me _with_ the spinOnce in the cancel

Then, can u approve this PR?  Also @spuetz, can you test and approve this?

As for the async spinner...  I see the point, but is a completely different issue and so we shouldn't mix with this PR

corot:
@dorezyuk , @spuetz , Description updated;  take a look at the :warning: section and tell me if u ratify my choice

mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
```diff
@@ -119,10 +119,12 @@ namespace mbf_abstract_nav
       double action_angle_tolerance = 3.1415);
 
     /**
-     * @brief Cancel the controller execution.
-     * This calls the cancel method of the controller plugin, sets the cancel_ flag to true,
-     * and waits for the control loop to stop. Normally called upon aborting the navigation.
-     * @return true, if the control loop stops within a cycle time.
+     * @brief Cancel the controller execution. Normally called upon aborting the navigation.
+     * This calls the cancel method of the controller plugin. If he answers true, he becomes responsible of
```
dorezyuk: if the plugins returns true

mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
```diff
@@ -119,10 +119,12 @@ namespace mbf_abstract_nav
...
+     * stopping, and we will keep requesting velocity commands until he returns CANCELED.
```
dorezyuk: it returns

mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
```diff
@@ -119,10 +119,12 @@ namespace mbf_abstract_nav
...
+     * If false, meaning cancel is not implemented, or that the controller defers handling it, MBF will set
```
dorezyuk: If it returns false

mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
```diff
@@ -119,10 +119,12 @@ namespace mbf_abstract_nav
...
+     * the cancel_ flag to true, and wait for the control loop to stop.
+     * @return true, if the controller handles stooping or the control loop stops within a cycle time.
```
dorezyuk: the stopping ... or if

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -254,18 +254,24 @@ bool AbstractControllerExecution::reachedGoalCheck()
 
 bool AbstractControllerExecution::cancel()
 {
-  // request the controller to cancel; it returns false if cancel is not implemented or rejected by the plugin
-  if (!controller_->cancel())
+  // Request the controller to cancel; returns true if he takes care of stopping, returning CANCELED when done,
```
dorezyuk: controller seem to be very male in this pr :grin:

mbf_abstract_nav/src/abstract_controller_execution.cpp
dorezyuk: any opinion here?

mbf_abstract_nav/src/abstract_controller_execution.cpp
corot: mmm....  I think I refer always as "him"...  :thinking: 
I'll start using it

mbf_abstract_nav/src/abstract_controller_execution.cpp
corot: but then on [this line](https://github.com/magazino/move_base_flex/pull/274/files#diff-bbabf6c9cc39ada410c0f58e40f9cf9c80a104f403ee9196f6a1519fe972a2dfL295) we will butch the smooth stopping, isn't it?

dorezyuk: ah right

dorezyuk (approved):

spuetz (approved):

Issue #287, opened 2021-12-23
multiple planner retries despite setting "planner_max_retries" to 0


ChristofDubs:
It seems that [this pull request](https://github.com/magazino/move_base_flex/pull/237) has introduced a bug:

When setting `planner_max_retries` to `0` and `planner_patience` to `>0`, in case `makePlan()` fails without exceeding the `planner_patience`, the last else case is reached:

https://github.com/magazino/move_base_flex/blob/da5d1e45acce55b1d92c672beec057dc44eda229/mbf_abstract_nav/src/abstract_planner_execution.cpp#L352-L355

so `makePlan()` will keep getting called. @dorezyuk actually pointed this out himself [in one of his comments](https://github.com/magazino/move_base_flex/pull/237#discussion_r516495028).

______________________________________________________________________________________________________

In terms of fixing this: It seems to me that for the check here

https://github.com/magazino/move_base_flex/blob/da5d1e45acce55b1d92c672beec057dc44eda229/mbf_abstract_nav/src/abstract_planner_execution.cpp#L332-L336

the initial version was correct, where it was `max_retries_ >= 0` instead?

ChristofDubs:
Alternative fix:

There is a test that checks that for `planner_max_retries = 0` and `planner_patience = 0` case, the return code is `NO_PLAN_FOUND` if `makePlan()` returns an error.

If that is the desired behavior, maybe we could instead consider removing the `&& patience_.isZero()` part here?

https://github.com/magazino/move_base_flex/blob/da5d1e45acce55b1d92c672beec057dc44eda229/mbf_abstract_nav/src/abstract_planner_execution.cpp#L347-L351

Doing so still makes the current tests pass.

dorezyuk:
Thx for reporting the issue.

If I set the patience to e.x. 1 sec (and max_retries to zero) the mbf tries to generate a path for 1 sec and fails after the patience elapses. To which value did you set the patience to get the infinite loop?

Maybe there is also a misunderstanding what should happen if max_retries is set to zero. My idea was to allow the user either to specify a number of retries - or if the number of retries is not interesting to allow him to disable the retries counting and just focus on the time spent in planning (or to specify both, max_retries and patience).

ChristofDubs:
> Thx for reporting the issue.
> 
> If I set the patience to e.x. 1 sec (and max_retries to zero) the mbf tries to generate a path for 1 sec and fails after the patience elapses. To which value did you set the patience to get the infinite loop?
[...]

PR #288, opened 2021-12-24
Fix multiple retries despite planner_max_retries=0 (#287)

ChristofDubs:
fix for #287

As described in the issue, if `planner_max_retries` is set to `0` and `makePlan()` fails before the patience is exceeded, the state should be set to `NO_PLAN_FOUND` immediately, and no retries should occur (just like in the case where `planner_patience` is set to 0).

I also added a test for this fix (and they fail without the fix), and took the opportunity to add a few other cases that didn't have tests yet.

### DIFF
--- a/mbf_abstract_nav/src/abstract_planner_execution.cpp
+++ b/mbf_abstract_nav/src/abstract_planner_execution.cpp
@@ -344,7 +344,7 @@ void AbstractPlannerExecution::run()
                                                 << (cancel_ ? "; planner canceled!" : ""));
           setState(PAT_EXCEEDED, true);
         }
-        else if (max_retries_ == 0 && patience_.isZero())
+        else if (max_retries_ == 0)
         {
           ROS_INFO_STREAM("Planning could not find a plan!");
           setState(NO_PLAN_FOUND, true);
--- a/mbf_abstract_nav/test/abstract_planner_execution.cpp
+++ b/mbf_abstract_nav/test/abstract_planner_execution.cpp
@@ -23,6 +23,7 @@ using mbf_abstract_nav::AbstractPlannerExecution;
 using mbf_abstract_nav::MoveBaseFlexConfig;
 using testing::_;
 using testing::AtLeast;
+using testing::InSequence;
 using testing::Return;
 using testing::Test;
 
@@ -125,8 +126,9 @@ TEST_F(AbstractPlannerExecutionFixture, success_after_retries)
 
   // setup the expectations
   AbstractPlannerMock& mock = dynamic_cast<AbstractPlannerMock&>(*planner_);
-  auto& exp = EXPECT_CALL(mock, makePlan(_, _, _, _, _, _)).Times(config.planner_max_retries).WillRepeatedly(Return(11));
-  EXPECT_CALL(mock, makePlan(_, _, _, _, _, _)).Times(1).After(exp).WillOnce(Return(1));
+  InSequence seq;
+  EXPECT_CALL(mock, makePlan(_, _, _, _, _, _)).Times(config.planner_max_retries).WillRepeatedly(Return(11));
+  EXPECT_CALL(mock, makePlan(_, _, _, _, _, _)).Times(1).WillOnce(Return(1));
 
   // call and wait
   ASSERT_TRUE(start(pose, pose, 0));
@@ -238,7 +240,8 @@ TEST_F(AbstractPlannerExecutionFixture, patience_exceeded_waiting_for_planner_re
 
 TEST_F(AbstractPlannerExecutionFixture, patience_exceeded_infinite_retries)
 {
-  // if negative retries are configured, we expect makePlan to repeatedly get called, and PAT_EXCEEDED to be returned once the patience is exceeded
+  // if negative retries are configured, we expect makePlan to repeatedly get called, and PAT_EXCEEDED to be returned
+  // once the patience is exceeded
 
   // configure the class
   MoveBaseFlexConfig config;
--- a/mbf_abstract_nav/test/abstract_planner_execution.cpp
+++ b/mbf_abstract_nav/test/abstract_planner_execution.cpp
@@ -22,6 +22,7 @@ struct AbstractPlannerMock : public AbstractPlanner
 using mbf_abstract_nav::AbstractPlannerExecution;
 using mbf_abstract_nav::MoveBaseFlexConfig;
 using testing::_;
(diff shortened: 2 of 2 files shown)


corot (approved):
lgtm;  I have added the explanation about how the different combinations on max_retries and patience work:
http://wiki.ros.org/move_base_flex#:~:text=mbf_simple_nav%20for%20details.-,Common%20Parameters,-~%3Cname%3E/planner_frequency%20(double

Actually, I have realized that for controller it works as without this PR!
https://github.com/magazino/move_base_flex/blob/53930c17632d45c0362eca14ca6a60cb83cc9ff2/mbf_abstract_nav/src/abstract_controller_execution.cpp#L415
I guess we should change it to make both match and also match to the documented behavior
@dorezyuk , @spuetz , do u agree?

dorezyuk:
Thanks for the tests and the pr.

I think the base logic we want to implement here is

| Patience  | Retries | Effect |On Failure|
| ------------- | ------------- |-|-|
| Enabled  | Enabled  |Both |PAT_EXCEEDED/MAX_RETRIES|
| Disabled  | Disabled  | No retries| NO_PLAN_FOUND|
|Enabled| Disabled| Patience|PAT_EXCEEDED|
|Disabled| Enabled| Retries|MAX_RETRIES|

The only question is if we say the "retry logic" is disabled at max_retries < 0 or max_retries == 0. Both is fine for me, but I thought it might be confusing if we disable the "patience logic" at zero and the "retry logic" at max_retries < 0. On one hand the max_retries < 0 would be more consistent with move_base. On the other hand - move_base does not support the disabling of the "patience logic" - so they don't have this inconsistency... But if you guys feel its more intuitive to disable the "retry logic" at max_retries < 0, I'm fine with that.

I don't entirely understand why we should return NO_PLAN_FOUND if we reach out max-retries (if its set to zero). For me this seems rather counter-intuitive.

The biggest problem with this pr is that it creates a dead lock for the second row in the table above. If we say "retry logic" is disabled at max_retries < 0 and if the also disable the patience (patience.isZero == True), the code will never exit the loop

ChristofDubs:
> The only question is if we say the "retry logic" is disabled at max_retries < 0 or max_retries == 0. Both is fine for me, but I thought it might be confusing if we disable the "patience logic" at zero and the "retry logic" at max_retries < 0. On one hand the max_retries < 0 would be more consistent with move_base. On the other hand - move_base does not support the disabling of the "patience logic" - so they don't have this inconsistency... But if you guys feel its more intuitive to disable the "retry logic" at max_retries < 0, I'm fine with that.

I'd be in favor of infinite attempts for the max_retries < 0 case:
- The user should be able to configure exactly one planning attempt (i.e. one attempt, no retries)
- I would expect that based on the parameter's name, `max_retries == 0` should mean `0` retries, which means one attempt and zero retries.

It seems that the parameter name max_**re**tries is what is causing the inconsistency to begin with: If it was instead called `max_attempts`, then `max_attempts = 1` would mean no retries (`max_retries == 0`) and `max_attempts = 0` would be infinite retries (`max_retries = -1`); this would then be consistent with patience disabling.

> I don't entirely understand why we should return NO_PLAN_FOUND if we reach out max-retries (if its set to zero). For me this seems rather counter-intuitive.

It just returns NO_PLAN_FOUND if it failed and it is not allowed to do any retries (which it already did before this PR); I was assuming that MAX_RETRIES should not be returned since it didn't do any retries:

| Patience | Retries | Effect | On Failure |
| --------------- | --------------- | --------------- |--------------- |
| Enabled	| > 0	 | Retries + Patience | PAT_EXCEEDED / MAX_RETRIES |
| Disabled | > 0 | Retries | MAX_RETRIES |
| Enabled	|  0	| 1 attempt + Patience | PAT_EXCEEDED / NO_PLAN_FOUND |
| Disabled |  0	| 1 attempt | NO_PLAN_FOUND |
| Enabled	|  < 0 | Patience | PAT_EXCEEDED |
| Disabled |  < 0 | infinite retries | - |

My personal opinion is that the two NO_PLAN_FOUND and MAX_RETRIES return codes are a redundant to begin with; I just based my changes on what was already there to not break any existing code.

Assuming we used NO_PLAN_FOUND regardless of how many retries were configured, then above table would simplify to:

| Patience | Retries | Effect | On Failure |
| --------------- | --------------- | --------------- |--------------- |
| Enabled	| >= 0 | finite attempts + Patience | PAT_EXCEEDED / NO_PLAN_FOUND |
| Disabled | >= 0 | finite attempts | NO_PLAN_FOUND |
| Enabled	|  < 0 | Patience | PAT_EXCEEDED |
| Disabled |  < 0 | infinite attempts | - |


> The biggest problem with this pr is that it creates a dead lock for the second row in the table above. If we say "retry logic" is disabled at max_retries < 0 and if the also disable the patience (patience.isZero == True), the code will never exit the loop

Setting `max_retries < 0` and `patience == 0` would already have this effect without this PR? So it's not a new problem that was introduced.

I also think this is a valid configuration, and exactly the behavior one would expect by giving infinite patience and infinite retires: It means the planner should never give up. I wouldn't recommend such a configuration, but it might be what the user wants if there are no alternative goals to target and no recovery behaviors. Or, the user has his own logic of when to cancel and re-send the action.

dorezyuk:
Ah ok, then the misunderstanding I had was that I considered the "infinite retries" as a bug.. So the question we should answer is if we want to allow something like a infinite retry for the user... @corot @spuetz What do you think?

corot:
> Ah ok, then the misunderstanding I had was that I considered the "infinite retries" as a bug.. So the question we should answer is if we want to allow something like a infinite retry for the user... @corot @spuetz What do you think?

Inf retries is a valid option;  :+1: for merging this as it is (and do the same for exe_path, see [my comment above](https://github.com/magazino/move_base_flex/pull/288#pullrequestreview-840016005))

spuetz:
> Ah ok, then the misunderstanding I had was that I considered the "infinite retries" as a bug.. So the question we should answer is if we want to allow something like a infinite retry for the user... @corot @spuetz What do you think?

Yes, inf retries as an option makes sense... as the action and by that the execution can be canceled.

corot:
> Yes, inf retries as an option makes sense... as the action and by that the execution can be canceled.

So we all agree with this solution?   if so,,, @spuetz, @dorezyuk, please approve

spuetz (approved):

dorezyuk (approved):

PR #289, opened 2022-01-18
Fix multiple retries despite controller_max_retries=0

corot:
As discussed [here](https://github.com/magazino/move_base_flex/pull/288#pullrequestreview-840016005), this PR replicates the current planner behavior when max_retries is zero, namely call controller only once.
This is also the behavior documented on [the wiki](http://wiki.ros.org/move_base_flex#:~:text=to%20be%20oscillating-,Note%20about%20max_retries%20and%20patience%20interaction%3A,-All%20possible%20combinations)

### DIFF
--- a/mbf_abstract_nav/src/abstract_controller_execution.cpp
+++ b/mbf_abstract_nav/src/abstract_controller_execution.cpp
@@ -411,8 +411,8 @@ void AbstractControllerExecution::run()
           else
           {
             setState(NO_LOCAL_CMD); // useful for server feedback
-            // we keep on moving if we have retries left or if the user has granted us some patience.
-            moving_ = max_retries_ || !patience_.isZero();
+            // keep trying if we have > 0 or -1 (infinite) retries
+            moving_ = max_retries_;
           }
           // could not compute a valid velocity command -> stop moving the robot
           publishZeroVelocity(); // command the robot to stop; we still feedback command calculated by the plugin


mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -411,8 +411,8 @@ void AbstractControllerExecution::run()
           else
           {
             setState(NO_LOCAL_CMD); // useful for server feedback
-            // we keep on moving if we have retries left or if the user has granted us some patience.
-            moving_ = max_retries_ || !patience_.isZero();
+            // keep trying if we have non-zero retries left
+            moving_ = max_retries_;
```
dorezyuk, line 415: so this will be casted to zero only if max_retries is zero and we will continue moving if max_retries is disabled (aka. negative)? I like!

dorezyuk: maybe add a comment that -1 is also casted to true (I had to double check that) or write `moving_ = max_retries != 0` so we express it clearer for people like me;)

dorezyuk (approved):

mbf_abstract_nav/src/abstract_controller_execution.cpp
corot: well,,,   -1 is also non-zero

spuetz: Could you explain why we should remove `!patience_.isZero();`?

spuetz: ping @corot  :)

corot: Before:  we keep retrying if retries is > 0 or we have not exhausted a *non-zero* patience
After:     we keep retrying if retries is > 0

That is, if retries is 0, we never retry, regardless of patience. That's what we have documented on the wiki

spuetz (approved):

PR #291, opened 2022-02-16
[fix] move_base_action: always sleep regardless of replanning

remco-r:
when `planner_frequency > 0` the replanning thread would never sleep and thus cause CPU load of 100%.
This fix makes sure there is always an `update_period.sleep()` in the loop.

This reduces the load of mbf_costmap_navigation to 8% in my case, without any different behavior.

### DIFF
--- a/mbf_abstract_nav/src/move_base_action.cpp
+++ b/mbf_abstract_nav/src/move_base_action.cpp
@@ -526,7 +526,7 @@ bool MoveBaseAction::replanningActive() const
 void MoveBaseAction::replanningThread()
 {
   ros::Duration update_period(0.005);
-  ros::Rate update_rate(200);
+  ros::Rate update_rate(1.0/update_period.toSec());
   ros::Time last_replan_time(0.0);
 
   while (ros::ok())
--- a/mbf_abstract_nav/src/move_base_action.cpp
+++ b/mbf_abstract_nav/src/move_base_action.cpp
@@ -526,6 +526,7 @@ bool MoveBaseAction::replanningActive() const
 void MoveBaseAction::replanningThread()
 {
   ros::Duration update_period(0.005);
+  ros::Rate update_rate(200);
   ros::Time last_replan_time(0.0);
 
   while (ros::ok())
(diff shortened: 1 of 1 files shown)


mbf_abstract_nav/src/move_base_action.cpp
```diff
@@ -553,16 +553,14 @@ void MoveBaseAction::replanningThread()
       }
       // else keep waiting for planning to complete (we already waited update_period in waitForResult)
     }
-    else if (!replanningActive())
-    {
-      update_period.sleep();
-    }
-    else if (ros::Time::now() - last_replan_time >= replanning_period_)
+
+    if (replanningActive() && ros::Time::now() - last_replan_time >= replanning_period_)
```
corot:
```suggestion
    else if (replanningActive() && ros::Time::now() - last_replan_time >= replanning_period_)
```

Otherwise we can send a new goal while the previous one is still running
A small issue with this is that `action_client_get_path_.getState().isDone()` can become true at this point, so we may delay the next planning cycle by `update_period`

corot: I can notice the big drop on CPU usage. The idea of the current implementation is that the [waitForResult](https://github.com/magazino/move_base_flex/pull/291/files#diff-6b8a12259c3d11b5f68c080c8a6d7a690aecea38015c66366f1d559b16bd4acbR535) makes unnecessary another sleep,  but looks like waitForResult is very wasteful!

Maybe a better implementation can be to use a ros::Rate: it waits a variable amount of time to fulfill a frequency. So it will automatically discount the time spent on waitForResult. Another advantage is that it will allow you to show a WARN whenever we cannot catch up with the configured replanning_period_.
Would you mind to re-implement with a Rate, @remco-r ?

corot (changes requested):

mbf_abstract_nav/src/move_base_action.cpp
remco-r: either waitForResult is wastefull, or most of the time it does not get there because `action_client_get_path_.getState().isDone()` is true. 
Either way, i have implemented your suggestions, and they work out well.

mbf_abstract_nav/src/move_base_action.cpp
```diff
@@ -526,6 +526,7 @@ bool MoveBaseAction::replanningActive() const
 void MoveBaseAction::replanningThread()
 {
   ros::Duration update_period(0.005);
+  ros::Rate update_rate(200);
```
corot: to avoid 2 const with the same value
```suggestion
  ros::Rate update_rate(1.0/update_period.toSec());
```

corot (approved):

mbf_abstract_nav/src/move_base_action.cpp
corot: Yep, waitForResult is very wasteful;  good to know!
Still, the original code was wrong also because we only waited when either planning was active or disabled, not when [waiting for the new replanning time](https://github.com/magazino/move_base_flex/pull/291/files#diff-6b8a12259c3d11b5f68c080c8a6d7a690aecea38015c66366f1d559b16bd4acbL560)

Issue #58, opened 2018-03-19
Force config. regeneration when adding new params


corot:
When adding new params to https://github.com/magazino/move_base_flex/blob/master/mbf_abstract_nav/src/mbf_abstract_nav/__init__.py, you need to touch https://github.com/magazino/move_base_flex/blob/master/mbf_costmap_nav/cfg/MoveBaseFlex.cfg before compiling so code for the new parameters is generated.

Cannot do better?

corot:
One ugly option is to always touch the .cfg file, but this makes compilation longer when only it's occasionally needed:

- file(TOUCH_NOCREATE cfg/MoveBaseFlex.cfg)    requires CMake 3.12
or
- execute_process(COMMAND touch cfg/MoveBaseFlex.cfg)

jspricke:
You would need to add the file as a dependency to the add_custom_command in the generate_dynamic_reconfigure_options macro, i.e. here:
https://github.com/ros/dynamic_reconfigure/blob/melodic-devel/cmake/dynamic_reconfigure-macros.cmake#L73
Maybe an extra parameter in for the macro works.
I didn't find a way to inject the dependency later.

corot:
> Maybe an extra parameter in for the macro works.

Unfortunately not; generate_dynamic_reconfigure_options tries to process it as any other .cfg file

jspricke:
>> Maybe an extra parameter in for the macro works.
>
>Unfortunately not; generate_dynamic_reconfigure_options tries to process it as any other .cfg file

I meant you can improve the generate_dynamic_reconfigure_options (i.e. 
patch it), such that it supports your use case.

corot:
Change PRequested. @jspricke, maybe you can comment it, if it's not described clearly enough.

PR #292, opened 2022-03-04
Handle infeasible global planner result when periodically replanning

remco-r:
While a goal is active, and a new goal is given, say through rviz, and this new goal is infeasible, then mbf aborts the active goal (good behavior).
While a goal is active, and through the `planner_frequency > 0` this goal is replanned and no feasible path is found, nothing happens (bad behavior). There will be no active global path, and the local planner will continue on its old path.

I connected the replanning goal call with the same callback as the "manual" goal call. so that the current goal will be canceled if no feasible path is found upon replanning.

### DIFF
--- a/mbf_abstract_nav/cfg/MoveBaseFlex.cfg
+++ b/mbf_abstract_nav/cfg/MoveBaseFlex.cfg
@@ -9,5 +9,6 @@ gen = ParameterGenerator()
 
 add_mbf_abstract_nav_params(gen)
 
+gen.add("stop_when_replan_fails", bool_t, 0, "Will cancel current goal when replanning fails", False)
 gen.add("restore_defaults", bool_t, 0, "Restore to the original configuration", False)
 exit(gen.generate(PACKAGE, "move_base_flex_node", "MoveBaseFlex"))
--- a/mbf_abstract_nav/cfg/MoveBaseFlex.cfg
+++ b/mbf_abstract_nav/cfg/MoveBaseFlex.cfg
@@ -9,6 +9,5 @@ gen = ParameterGenerator()
 
 add_mbf_abstract_nav_params(gen)
 
-gen.add("stop_when_replan_fails", bool_t, 0, "Will cancel current goal when replanning fails", False)
 gen.add("restore_defaults", bool_t, 0, "Restore to the original configuration", False)
 exit(gen.generate(PACKAGE, "move_base_flex_node", "MoveBaseFlex"))
--- a/mbf_abstract_nav/include/mbf_abstract_nav/move_base_action.h
+++ b/mbf_abstract_nav/include/mbf_abstract_nav/move_base_action.h
@@ -170,6 +170,9 @@ class MoveBaseAction
   //! true, if recovery behavior for the MoveBase action is enabled.
   bool recovery_enabled_;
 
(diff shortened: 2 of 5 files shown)


corot:
Not 100% sure this is a good idea  :thinking: 
Replanning can fail sporadically for whatever reason, but if the robot can keep moving towards the goal, it's better not to interrupt him, I guess.

remco-r:
It's indeed a matter of strategy. 

1. If your robot currently has only one goal, perhaps it should get as close as possible to it, even when sporadically no feasible global path can be found, or when there is actually no feasible path. 

2. However, when going over a list of goals you may want to continue with the next one if you know that the current goal is infeasible, in stead of getting closer until the local planner finds out that its infeasible.

The question is: When replanning active, do you want to respond to the latest global planner run, or do you want to follow the last _feasible_ path?

I must say i agree with you and lean towards 1. a bit more.
Another option would be to create a parameter for this, something like `handle_infeasible_replanning_outcome`.
I leave it up to you.

corot:
> I must say i agree with you and lean towards 1. a bit more. Another option would be to create a parameter for this, something like `handle_infeasible_replanning_outcome`. I leave it up to you.

yup, I was thinking about having the change under a parameter;  I propose a more straightforward name, like `stop_on_replan_failure`, or `stop_when_replan_fails`

remco-r:
implemented the `stop_when_replan_fails` as dynamic reconfigure param. However I'm having issues showing it in my rqt_reconfigure, so i only tested it by hard coding.

Also, i noticed `replanningThread` is a separate thread, which does access some member variables. I would be willing to make those atomic.

corot (approved):
I made another commit to move stop_when_replan_fails to common abstract config, hope u don't mind the meddling.
The reason is explained [here](https://github.com/magazino/move_base_flex/pull/292/files#diff-0fc00860ec919c88d25ae0e61193d3d908c20370e7b8bcdceac242aab5f9da40R1-R11)

The need to copy manually all the common params is a bit crappy, but at the time I didn't find a better solution

And as for triggering the recompilation, I created an issue requesting help, w/o much success :crying_cat_face: 
https://github.com/magazino/move_base_flex/issues/58

remco-r:
That's fine. explanation link doesn't work though. 

Great to know about the cfg recompilation issues. I was struggling with that.

corot:
> That's fine. explanation link doesn't work though.
> 
> Great to know about the cfg recompilation issues. I was struggling with that.

Link is to this file: `mbf_abstract_nav/src/mbf_abstract_nav/__init__.py`, more precisely this comment:

```
# Generic set of parameters required by any MBF-based navigation framework
# To use:
#
#  from mbf_abstract_nav import add_mbf_abstract_nav_params
#  gen = ParameterGenerator()
#  add_mbf_abstract_nav_params(gen)
#  ...
# WARN: parameters added here must be copied on the specific MBF implementation reconfigure callback, e.g. in:
#       https://github.com/magazino/move_base_flex/blob/master/mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp#L130
# Also, you need to touch https://github.com/magazino/move_base_flex/blob/master/mbf_costmap_nav/cfg/MoveBaseFlex.cfg
# when recompiling to ensure configuration code is regenerated with the new parameters

Issue #184, opened 2020-04-21
Remove dependency on base_local_planner


spuetz:
For release purposes, we should remove the dependency on the `base_local_planner` in the `mbf_costmap_nav` package.
I would suggest to simply copy the necessary code to a mbf_costmap_util folder, which can be installed as separate library.

This enable us to enhance the existing move_base plugins with our interfaces.

corot:
:thinking:  what's the problem of having `base_local_planner` as dependency?

spuetz:
Circular dependency if we implement the new interface to the plugins.

corot:
Addressed on #186

PR #293, opened 2022-03-07
Implement getRobotVelocity on RobotInformation and use on executions instead of tf

corot:
How can I run locally run-style-check-diff.sh so I format only the modified lines?

Our system is a bit cumbersome... as I need to format just the modified lines :thinking: 
So I'll let the previous format


mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
```diff
@@ -220,11 +223,11 @@ namespace mbf_abstract_nav
...
     std::string plugin_name_;
 
-    //! the local planer to calculate the velocity command
+    //! The local planer to calculate the velocity command
     mbf_abstract_core::AbstractController::Ptr controller_;
 
-    //! shared pointer to the shared tf listener
-    const TFPtr &tf_listener_ptr;
+    //! Reference to the current robot state
+    mbf_utility::RobotInformation &robot_info_;
```
corot: It should be a const reference, but bloody OdometryHelper doesn't declare its methods as const
Options:
- PR OdometryHelper: will take ages to merge, if ever
- Copy it onto mbf_utility and modify
- Let it this way

If nothing against, I'l go for the 1st and 2nd options, on a separated commit for easy reverting, in case 1st gets merged fast.
Sounds good?  @spuetz , @dorezyuk

dorezyuk: A better solution is to declare the odom helper in the RobotInformation as mutable (and thus make the methods using the odom helper const)

corot: ah...  good idea!  changed

mbf_utility/src/robot_information.cpp
```diff
@@ -65,10 +69,20 @@ bool RobotInformation::getRobotPose(geometry_msgs::PoseStamped &robot_pose) cons
   return true;
 }
 
-bool RobotInformation::getRobotVelocity(geometry_msgs::TwistStamped &robot_velocity, ros::Duration look_back_duration) const
+bool RobotInformation::getRobotVelocity(geometry_msgs::TwistStamped &robot_velocity) const
+{
+  nav_msgs::Odometry base_odom;
+  odom_helper_.getOdom(base_odom);
```
dorezyuk, line 79: would it make sense to check here the data? something like 
```c++
if(!base_odom.header.seq && base_odom.header.stamp == ros::Time(0)){
  ROS_WARN("No odometry message received");
  return false;
}
```

mbf_utility/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ cmake_minimum_required(VERSION 3.0.2)
 project(mbf_utility)
 
 find_package(catkin REQUIRED COMPONENTS
+  base_local_planner
```
dorezyuk: Didn't we had some issues with this dependency? Or why did we copy the footprint helper?

mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
```diff
@@ -82,16 +81,19 @@ namespace mbf_abstract_nav
...
+     * @param controller_ptr Pointer to the controller plugin
+     * @param robot_info Current robot state
+     * @param vel_pub Velocity publisher
+     * @param goal_pub Current goal publisher
+     * @param config Initial configuration for this execution
      */
     AbstractControllerExecution(
         const std::string &name,
         const mbf_abstract_core::AbstractController::Ptr &controller_ptr,
+        const mbf_utility::RobotInformation &robot_info,
```
dorezyuk, line 94: I think this one is missing in the includes

corot: ah... shit, I forgot to update the tests :see_no_evil:

mbf_utility/CMakeLists.txt
corot: yup.....  https://github.com/magazino/move_base_flex/issues/184

mbf_utility/src/robot_information.cpp
corot: yes, I thought about that, but... we cannot warranty that odom is properly stamped, so I didn't take the risk

but looks like we must copy odom helper (see below), so I can do a proper implementation of this check

corot:
About the formatting....  are u ok with an extra commit formatting to clang the modified code?
some files have wrong indentation, so changes will look bigger than what they really are

Maybe I can do after having the approvals

mbf_utility/src/robot_information.cpp
corot: now implemented

mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
corot: coopted odom helper

corot:
Ok, I coopted odom helper. Ready for a final review, @dorezyuk, @spuetz 

And let me know what I do with the formatting;  I have the commit waiting for after your reviews

spuetz:
Yea please change the formatting back. Formatting all the other files should be in a separate PR. I hope this is not too much work. Cheers - Seb

corot:
> Yea please change the formatting back. Formatting all the other files should be in a separate PR. I hope this is not too much work. Cheers - Seb

No, no,,, I didn't format anything on the PR.   I was just explaining why the check fails and offering to address that

mbf_abstract_nav/test/abstract_controller_execution.cpp
```diff
@@ -180,7 +181,7 @@ TEST_F(ComputeRobotPoseFixture, arrivedGoal)
   plan.back().header.frame_id = global_frame_;
   plan.back().pose.orientation.w = 1;
 
-  // make the toleraces small
+  // make the tolerates small
```
ChristofDubs: I suppose you meant `tolerances` ?

mbf_utility/src/robot_information.cpp
```diff
@@ -59,16 +61,33 @@ bool RobotInformation::getRobotPose(geometry_msgs::PoseStamped &robot_pose) cons
...
-bool RobotInformation::getRobotVelocity(geometry_msgs::TwistStamped &robot_velocity, ros::Duration look_back_duration) const
+bool RobotInformation::getRobotVelocity(geometry_msgs::TwistStamped &robot_velocity) const
 {
-  // TODO implement and filter tf data to compute velocity.
-  return false;
+  nav_msgs::Odometry base_odom;
+  odom_helper_.getOdom(base_odom);
+  if (base_odom.header.stamp.isZero())
+  {
+    ROS_ERROR_STREAM("No odometry messages received; robot velocity unknown");
```
ChristofDubs: I think there is some room for improvement here:
- given that user code worked so far without an `odom` topic being present, this should be a warning rather than an error?
- since this is not expected to spuriously fail / if it fails, it will most likely continue to fail, use a `THROTTLE` version
- add the odom topic to the warning using `base_odom.getOdomTopic()`

On the larger picture, it seems the `RobotInformation` class in `AbstractNavigationServer` used the default parameter value of the odom topic. Would be better to expose this as a ROS parameter (default initialized to `"odom"`)? Then, the warning here could be extended to inform the user to set this parameter. (Maybe could even allow the user to suppress the warning by explicitly setting the odom topic parameter to an empty string)

mbf_utility/include/mbf_utility/robot_information.h
```diff
@@ -69,7 +71,20 @@ class RobotInformation
...
+   */
+  bool getRobotVelocity(geometry_msgs::TwistStamped &robot_velocity) const;
+
+  /**
+   * @brief Check whether the robot is stopped or not
+   * @param rot_stopped_velocity The rotational velocity below which the robot is considered stopped
+   * @param trans_stopped_velocity The translational velocity below which the robot is considered stopped
+   * @return true if the robot is stopped, false otherwise
+   */
+  bool isRobotStopped(double rot_stopped_velocity, double trans_stopped_velocity) const;
```
ChristofDubs, line 87: this is unused, is that intentional?

corot: Yes; we want to replace base_local_planner with the same functionality, and it provides this check

mbf_utility/src/robot_information.cpp
corot: good idea

ChristofDubs (approved):

PR #296, opened 2022-03-10
Fix race condition between finishing and updating a goal

### LINKED ISSUE #295: Race condition when updating goal for ExePath
The following is run in a thread

```
//AbstractActionBase
  virtual void run(ConcurrencySlot &slot)
  {
    slot.execution->preRun();
    runImpl(slot.goal_handle, *slot.execution);
    ROS_DEBUG_STREAM_NAMED(name_, "Finished action \"" << name_ << "\" run method, waiting for execution thread to finish.");
    slot.execution->join();
    ROS_DEBUG_STREAM_NAMED(name_, "Execution completed with goal status "
                           << (int)slot.goal_handle.getGoalStatus().status << ": "<< slot.goal_handle.getGoalStatus().text);
    slot.execution->postRun();
    slot.in_use = false;
  }
```
`ControllerAction::start` uses the boolean `slot.in_use` to see that the thread is still running.
see https://github.com/magazino/move_base_flex/blob/master/mbf_abstract_nav/src/controller_action.cpp#L69
But it is used to infer whether we are still in runImpl, i.e., the goal is active and the exe_path controller is processing it.

The race condition is the following sequence:
- runImpl is done
  - Set also the goal to succeeded, for example, because goal is reached
- Another goal comes in, we call `ControllerAction::start` for it
- `ControllerAction::start` sees `slot.in_use==true`
- `ControllerAction::start` wants to cancel the goal that was succeeded, yields
  - `[/navigation/move_base_flex ERROR 1646052687.690027]: To transition to a cancelled state, the goal must be in a pending, recalling, active, or preempting state, it is currently in state: 3`
  - We update the internal structure for the thread about to terminate, and critical, also we accept the new goal but no one is going to process the goal
  - If I understand the logic in `AbstractActionBase::start` this goal is then also never cancelled upon a potential new goal arriving.
- Run sets now `slot.in_use=false`

Unfortunately, the node was not running with a log-level of debug when we hit the race condition. So I do not have the log about "Finished action ...".

RainerKuemmerle:
AbstractActionBase sets slot.in_use to false after having left runImpl.
This means that a check based on slot.in_use might still see a thread as
running despite the goal got assigned a terminal state.
Hence, we also check that the goal is still active to consume an update.

Fix #295

### DIFF
--- a/mbf_abstract_nav/src/controller_action.cpp
+++ b/mbf_abstract_nav/src/controller_action.cpp
@@ -69,9 +69,11 @@ void ControllerAction::start(
   if(slot_it != concurrency_slots_.end() && slot_it->second.in_use)
   {
     boost::lock_guard<boost::mutex> goal_guard(goal_mtx_);
-    if(slot_it->second.execution->getName() == goal_handle.getGoal()->controller ||
-       goal_handle.getGoal()->controller.empty())
+    if ((slot_it->second.execution->getName() == goal_handle.getGoal()->controller ||
+         goal_handle.getGoal()->controller.empty()) &&
+        slot_it->second.goal_handle.getGoalStatus().status == actionlib_msgs::GoalStatus::ACTIVE)
     {
+      ROS_DEBUG_STREAM_NAMED(name_, "Updating running controller goal of slot " << static_cast<int>(slot));
       update_plan = true;
       // Goal requests to run the same controller on the same concurrency slot already in use:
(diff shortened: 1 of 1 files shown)


corot (approved):
I didn't try to reproduce the race condition (I guess it won't be easy), but the change makes full sense

spuetz (approved):

Eman7C7 (approved):

PR #304, opened 2022-09-09
Add error ranges for all four actions

corot:
a bit too late to do this, but,,, we should have done from the beginning!

### DIFF
--- a/mbf_msgs/action/ExePath.action
+++ b/mbf_msgs/action/ExePath.action
@@ -15,29 +15,31 @@ float32 angle_tolerance
 ---
 
 # Predefined success codes:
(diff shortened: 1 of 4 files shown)


ChristofDubs (approved):

RainerKuemmerle (approved):

tomin11 (approved):

spuetz:
This will rebuild the messages and break the binary compatibility as it checks the checksum of the received messages. But just for the master branch this should be fine. For newer releases in the direction to ROS 2 this makes sense. I'll approve this, too

spuetz (approved):

PR #306, opened 2022-12-05
Add parameter force_stop_on_retry that avoids zero vel commands on recoverable errors

corot:
This allows gracefully slowing down while retrying calls to the controller.
Note that this will apply only as long as 
- we have not exhausted available controller retries 
- we have not exhausted patience time

When either condition doesn't hold, we send zero velocity commands, as with current implementation

Default value is true, meaning current behavior

### DIFF
--- a/mbf_abstract_nav/src/abstract_controller_execution.cpp
+++ b/mbf_abstract_nav/src/abstract_controller_execution.cpp
@@ -66,6 +66,7 @@ AbstractControllerExecution::AbstractControllerExecution(
   private_nh.param("robot_frame", robot_frame_, std::string("base_link"));
   private_nh.param("map_frame", global_frame_, std::string("map"));
   private_nh.param("force_stop_at_goal", force_stop_at_goal_, false);
+  private_nh.param("force_stop_on_retry", force_stop_on_retry_, true);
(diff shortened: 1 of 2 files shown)


mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -401,8 +401,18 @@ void AbstractControllerExecution::run()
             // keep trying if we have > 0 or -1 (infinite) retries
             moving_ = max_retries_;
           }
-          // could not compute a valid velocity command -> stop moving the robot
-          publishZeroVelocity(); // command the robot to stop; we still feedback command calculated by the plugin
+
+          // could not compute a valid velocity command
+          if (!moving_ || force_stop_on_retry_)
```
RainerKuemmerle, line 407: We seem to miss reading a param for force_stop_on_retry_ - making it uninitialized? Probably the default value should be true to keep the old behavior?

corot: :see_no_evil:     thanks

I copy-pasted the changes from the robot to push, and missed this one

RainerKuemmerle (approved):

ChristofDubs (approved):

Issue #297, opened 2022-03-22
Costmap Safety Check Doesn't Quite Work as Expected


Jmeyer1292:
Greetings. MoveBaseFlex is an excellent piece of software. Thank you for the hard work in creating and maintaining it.

I wanted to report an issue: In brief the `safetyChecks` are not working as expected.

 * **Setup:** Robot with MoveBaseFlex and a Time-Elastic-Band local controller
 * **Experiment**: Set an expected sensor data rate in your costmap parameters. Do NOT publish sensor data to the costmap.
 * **Expectation**: The robot does not move. If sensor data is provided, it would keep going.
 * **Observation**: The robot stutters and slowly creeps forward. See figure 1 for a visualization of the bag file.
  
Blue is a velocity command (`linear/x`).
Red is measured robot twist (`linear/x`). 
![oscillation](https://user-images.githubusercontent.com/4960180/159584473-4c38fce5-2d4b-413a-9b53-3b69dd967846.png)
**Figure 1**

A very quick glance through the source code shows [this](https://github.com/magazino/move_base_flex/blob/e70201c7d206971817bec50a5a9b037ef21a7e94/mbf_abstract_nav/src/abstract_controller_execution.cpp#L318) section of code as the point where our checks are applied.

It _appears_ that the safety check fails (as expected), a zero velocity is published, and the thread sleeps. It then proceeds to go and run the controller and issue a command anyway. I suggest the inclusion of a `continue` in the loop where the safety checks fail.

I'm happy to make this edit, but I acknowledge that this may have been done on purpose. I can certainly see the value in still running controllers at regular intervals and just letting them know that things are kinda broken.

Again thank you for being open source maintainers.

corot:
Thanks for the thorough analysis. I agree, at first glance, looks like a `continue;` is missing.
I'll try myself and see if there's any good reason for not having it (we wrote that logic long time ago, and I don't remember the rationale)

PR #307, opened 2022-12-05
Issue #297: Send just zero velocity commands when safety check doesn't pass

corot:
Thanks to @Jmeyer1292 for the investigation and simple solution

### DIFF
--- a/mbf_abstract_nav/src/abstract_controller_execution.cpp
+++ b/mbf_abstract_nav/src/abstract_controller_execution.cpp
@@ -298,8 +298,9 @@ void AbstractControllerExecution::run()
       {
         // the specific implementation must have detected a risk situation; at this abstract level, we
         // cannot tell what the problem is, but anyway we command the robot to stop to avoid crashes
-        publishZeroVelocity();   // note that we still feedback the command calculated by the plugin
+        publishZeroVelocity();
         loop_rate_.sleep();
+        continue;
       }
 
       // update plan dynamically


mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -298,8 +298,9 @@ void AbstractControllerExecution::run()
       {
         // the specific implementation must have detected a risk situation; at this abstract level, we
         // cannot tell what the problem is, but anyway we command the robot to stop to avoid crashes
-        publishZeroVelocity();   // note that we still feedback the command calculated by the plugin
+        publishZeroVelocity();
         loop_rate_.sleep();
+        continue;
```
ChristofDubs, line 303: Clarification question: If this safety check keeps failing, is there some mechanism that interrupts this at some point (by e.g. setting `cancel_` to true)?

corot: Yes, canceling the goal will interrupt this
But maybe we should add here a call to `isPatienceExceeded` and fail goal if it returns true?
:thinking:

ChristofDubs: Yeah that would be an improvement. However, the patience is currently allocated solely for the controller, so its meaning would change a bit, and the corresponding message in `isPatienceExceeded()` would need updating:

https://github.com/magazino/move_base_flex/blob/af3afbf37caa17a44ff41c71f8d1b3cc658455a2/mbf_abstract_nav/src/abstract_controller_execution.cpp#L212

renan028:
> Thanks to @Jmeyer1292 for the investigation and simple solution

tested:

before:

https://user-images.githubusercontent.com/6097267/228436790-165e5b73-83a7-44cf-a39e-d184eccb81fd.mp4

after:

https://user-images.githubusercontent.com/6097267/228436549-aee62d88-5ec2-45c4-806e-11311bbf2721.mp4

mbf_abstract_nav/src/abstract_controller_execution.cpp
corot: Or maybe moving the check to somewhere after checking patience and retries?

As it is clearly a bug, I'll merge this as it is by now and investigate on another PR how to react when this happens
Depending on the use case, it can be that we don't want to fail while the safety check is false for a relatively long time.
e.g. someone is using this to force the robot to wait till another agent crosses, or an elevator opens,,,

ChristofDubs: ok sure

ChristofDubs (approved):

PR #309, opened 2022-12-12
add BLOCKED_GOAL error code to ExePath action

ChristofDubs:
Add error code `BLOCKED_GOAL` to the ExePath action, such that the controller can report that the goal is blocked.

This seems cleaner than using the already existing `GOAL_BLOCKED   = 15` error code in [MoveBaseAction](https://github.com/magazino/move_base_flex/blob/master/mbf_msgs/action/MoveBase.action#L26) for this scenario since the new error code is within the `[ExePathResult::ERROR_RANGE_START, ExePathResult::ERROR_RANGE_END]` range.


corot:
> This seems cleaner than using the already existing `GOAL_BLOCKED = 15` error code in [MoveBaseAction](https://github.com/magazino/move_base_flex/blob/master/mbf_msgs/action/MoveBase.action#L26) for this scenario since the new error code is within the `[ExePathResult::ERROR_RANGE_START, ExePathResult::ERROR_RANGE_END]` range.

I think it's fine to have the same error repeated on more than one range, as this conveys at which level we detect the error condition

corot (approved):

spuetz (approved):

siferati (commented):
Q: The message definition already allows the controller to return any custom errors it needs between 121 and 149. Are we going to change the message definition every time we need a new error code...?

https://github.com/magazino/move_base_flex/blob/971832a74ec0ade0b0fdd5b1b293139b5ae0214c/mbf_msgs/action/ExePath.action#L44

F: Need to update `mbf_utility::outcome2str` with the new error code

https://github.com/magazino/move_base_flex/blob/af3afbf37caa17a44ff41c71f8d1b3cc658455a2/mbf_utility/src/navigation_utility.cpp#L212

ChristofDubs:
> Q: The message definition already allows the controller to return any custom errors it needs between 121 and 149. Are we going to change the message definition every time we need a new error code...?

Well it's a judgement call whether an error code is general enough to go into the 100 to 120 range; and in the case of `BLOCKED_GOAL`, that seems to be a general scenario that is not a plugin specific error.

> F: Need to update `mbf_utility::outcome2str` with the new error code

Yes, added in 3b9f8b697750d274584d5f04cd1a87a857d4fab8

siferati (approved):

PR #310, opened 2022-12-13
Add plugin-specific error ranges, so we don't return "unknown error"

corot:
Follow up of #304, with ranges for the plugin-specific error codes

rationale: It doesn't make sense to return "unknown error code" for codes reserved for the plugins

### DIFF
--- a/mbf_utility/src/navigation_utility.cpp
+++ b/mbf_utility/src/navigation_utility.cpp
@@ -211,65 +211,122 @@ double angle(const geometry_msgs::PoseStamped &pose1, const geometry_msgs::PoseS
 
 std::string outcome2str(unsigned int outcome)
 {
-  switch (outcome)
-  {
-    case mbf_msgs::MoveBaseResult::SUCCESS:         return "Success";
-    case mbf_msgs::MoveBaseResult::FAILURE:         return "Failure";
-    case mbf_msgs::MoveBaseResult::CANCELED:        return "Canceled";
-    case mbf_msgs::MoveBaseResult::COLLISION:       return "Collision";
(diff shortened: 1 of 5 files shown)


mbf_utility/src/navigation_utility.cpp
```diff
@@ -211,65 +211,122 @@ double angle(const geometry_msgs::PoseStamped &pose1, const geometry_msgs::PoseS
...
+    return "Not initialized";
+  if (outcome == mbf_msgs::RecoveryResult::INVALID_PLUGIN)
+    return "Invalid Plugin";
+  if (outcome == mbf_msgs::RecoveryResult::INTERNAL_ERROR)
+    return "Internal Error";
+  if (outcome == mbf_msgs::RecoveryResult::IMPASSABLE)
+    return "Impassable";
+  if (outcome == mbf_msgs::RecoveryResult::STOPPED)
+    return "Stopped";
+  if (outcome >= mbf_msgs::RecoveryResult::PLUGIN_ERROR_RANGE_START &&
```
renan028, line 325: S: keep the switch case, add this if at the start.
or, if you are ok of printing the enum as string (e.g. when you get `mbf_msgs::RecoveryResult::FAILURE` the output would be `mbf_msgs::RecoveryResult::FAILURE` as string), you can use magic_enum:
```
magic_enum::enum_name(error));
```

corot: I wanted to keep codes grouped by action, that's why I removed the switch

renan028 (approved):

ChristofDubs (approved):

PR #314, opened 2023-02-24
fix(mbf_abstract_nav): check for empty path to prevent segfault

mbeutelspacher:
Otherwise https://github.com/magazino/move_base_flex/blob/master/mbf_abstract_nav/src/controller_action.cpp#L87 will segfault when the path is empty

### DIFF
--- a/mbf_abstract_nav/src/controller_action.cpp
+++ b/mbf_abstract_nav/src/controller_action.cpp
@@ -63,7 +63,7 @@ void ControllerAction::start(
   if (goal_handle.getGoal()->path.poses.empty())
   {
     mbf_msgs::ExePathResult result;
-    result.message = "Path is empty.";
+    result.message = "Path is empty";
     result.outcome = mbf_msgs::ExePathResult::INVALID_PATH;
     goal_handle.setRejected(result, result.message);
   }
--- a/mbf_abstract_nav/src/controller_action.cpp
+++ b/mbf_abstract_nav/src/controller_action.cpp
@@ -60,6 +60,13 @@ void ControllerAction::start(
     goal_handle.setCanceled();
     return;
   }
+  if (goal_handle.getGoal()->path.poses.empty())
+  {
+    mbf_msgs::ExePathResult result;
+    result.message = "Path is empty.";
+    result.outcome = mbf_msgs::ExePathResult::INVALID_PATH;
+    goal_handle.setRejected(result, result.message);
+  }
 
   uint8_t slot = goal_handle.getGoal()->concurrency_slot;
 


mbf_abstract_nav/src/controller_action.cpp
```diff
@@ -60,6 +60,13 @@ void ControllerAction::start(
     goal_handle.setCanceled();
     return;
   }
+  if (goal_handle.getGoal()->path.poses.empty())
+  {
+    mbf_msgs::ExePathResult result;
+    result.message = "Path is empty.";
```
corot: for consistency
```suggestion
    result.message = "Path is empty";
```

mbf_abstract_nav/src/controller_action.cpp
```diff
@@ -60,6 +60,13 @@ void ControllerAction::start(
     goal_handle.setCanceled();
     return;
   }
+  if (goal_handle.getGoal()->path.poses.empty())
```
corot, line 63: Maybe a bit contrived, but... Couldn't be the case that someone uses an empty Path msg with a frame set on the header to command a robot to reach a tf frame?
:thinking: 
Just a thought that came to my mind

mbeutelspacher: Wouldn't you then rather send a path one pose ( the zero pose) in that frame? We could provide the "syntactic sugar" to automatically add this in case one sends an empty path but is this really expected behavior? If one sends an empty path in the global frame and the robots drives to map origin without any notice why he does that?

corot: mmm... true; well, I said it was a contrived case.
But I checked and MBF doesn't crash with an empty path. I guess must be the particular controller you use;  can be?
If so, I would add the check at controller level and let MBF simply report back the error, possibly as an [INVALID_PATH error ](https://github.com/magazino/move_base_flex/blob/master/mbf_msgs/action/ExePath.action#L33)

mbeutelspacher: it's crashing for in line 87 of this very file here (as mentioned above) , not sure how this can be controller dependent, but in my opinion that should be fixed here. I'll try to provide a regression test.

mbeutelspacher: So for me it crashes if the controller has an active goal, and you then call exe_path with an empty band.

mbeutelspacher: but yes for consistency we should probably call setAborted here, as we do when the controller is not executing a goal at the moment

corot:
@mbeutelspacher, did you find out why MBF was crashing in your case?
For the reasons we discussed above, I feel inclined to close this PR, but always thanks for the contribution!

PR #316, opened 2023-03-08
Create new service FindValidPose

corot:
New MBF service to add to mbf_costmap_nav: it goes in the line of the suggestion at issue #8; but less ambitious by now.
But we can add other services that for example finds a valid pose within an area.

The implementation is on PR #317

### DIFF
--- a/mbf_msgs/srv/CheckPath.srv
+++ b/mbf_msgs/srv/CheckPath.srv
@@ -21,6 +21,6 @@ uint8                      UNKNOWN   =  3    # path is partially in unknown spac
 uint8                      OUTSIDE   =  4    # path is partially outside the map
 
 uint32                     last_checked      # index of the first pose along the path with return_on state or worse
-uint8                      state             # path worst state (until last_checked): FREE, INFLATED, LETHAL, UNKNOWN...
+uint8                      state             # path worst state (until last_checked): FREE, INSCRIBED, LETHAL, UNKNOWN...
 uint32                     cost              # cost of all cells traversed by path within footprint padded by safety_dist
                                              # or just by the path if path_cells_only is true
--- a/mbf_msgs/srv/FindValidPose.srv
+++ b/mbf_msgs/srv/FindValidPose.srv
@@ -1,0 +1,27 @@
+# Find the closest valid pose to the given one, within the given linear and angular tolerances.
+# A pose is considered valid if the robot footprint there is entirely inside the map and not in
+# collision nor unknown space.
+# Alternatively this service can be use to find the closest valid pose to the robot's current one.
+
+uint8                      LOCAL_COSTMAP  = 1
+uint8                      GLOBAL_COSTMAP = 2
+
+geometry_msgs/PoseStamped  pose              # the starting pose from which we start the search
+float32                    safety_dist       # minimum distance allowed to the closest obstacle
+float32                    dist_tolerance    # maximum distance we can deviate from the given pose during the search
+float32                    angle_tolerance   # maximum angle we can rotate the given pose during the search
+uint8                      costmap           # costmap in which to check the pose
+bool                       current_pose      # check current robot pose instead (ignores pose field)
+bool                       use_padded_fp     # include footprint padding when checking cost; note that safety distance
+                                             # will be measured from the padded footprint
+---
+uint8                      FREE      =  0    # found pose is completely in traversable space
+uint8                      INSCRIBED =  1    # found pose is partially in inscribed space
+uint8                      LETHAL    =  2    # found pose is partially in collision
+uint8                      UNKNOWN   =  3    # found pose is partially in unknown space
+uint8                      OUTSIDE   =  4    # found pose is partially outside the map
+
+bool                       valid             # whether we found a valid pose
+uint8                      state             # found pose's state: FREE, INFLATED, LETHAL, UNKNOWN or OUTSIDE
+uint32                     cost              # found pose's cost (added cost of all cells covered by the footprint)
+geometry_msgs/PoseStamped  pose              # the pose found (filled only if valid is true)
--- a/mbf_msgs/srv/FindValidPose.srv
+++ b/mbf_msgs/srv/FindValidPose.srv
@@ -1,7 +1,13 @@
-# Find the closest valid pose to the given one, within the given linear and angular tolerances.
-# A pose is considered valid if the robot footprint there is entirely inside the map and not in
+# Find the closest free pose to the given one, within the given linear and angular tolerances.
+#
+# A pose is considered free if the robot footprint there is entirely inside the map and not in
 # collision nor unknown space.
-# Alternatively this service can be use to find the closest valid pose to the robot's current one.
+#
+# If no free pose can be found, but we find one partially on unknown space, or outside the map,
+# we will return it and set state to the correspondent option (unknown space takes precedence).
+# Otherwise state will be set to LETHAL.
+#
+# You can also instruct this service to use current robot's pose, instead of providing one.
 
 uint8                      LOCAL_COSTMAP  = 1
 uint8                      GLOBAL_COSTMAP = 2
@@ -23,4 +29,4 @@ uint8                      OUTSIDE   =  4    # found pose is partially outside t
 
 uint8                      state             # found pose's state: FREE, INSCRIBED, LETHAL, UNKNOWN or OUTSIDE
 uint32                     cost              # found pose's cost (added cost of all cells covered by the footprint)
-geometry_msgs/PoseStamped  pose              # the pose found (filled only if valid is true)
+geometry_msgs/PoseStamped  pose              # the pose found (filled only if state is not set to LETHAL)
--- a/mbf_msgs/srv/FindValidPose.srv
+++ b/mbf_msgs/srv/FindValidPose.srv
@@ -1,6 +1,6 @@
 # Find the closest free pose to the given one, within the given linear and angular tolerances.
 #
-# A pose is considered free if the robot footprint there is entirely inside the map and not in
+# A pose is considered free if the robot footprint there is entirely inside the map and neither in
 # collision nor unknown space.
 #
 # If no free pose can be found, but we find one partially on unknown space, or outside the map,
--- a/mbf_msgs/srv/FindValidPose.srv
+++ b/mbf_msgs/srv/FindValidPose.srv
@@ -21,7 +21,6 @@ uint8                      LETHAL    =  2    # found pose is partially in collis
 uint8                      UNKNOWN   =  3    # found pose is partially in unknown space
 uint8                      OUTSIDE   =  4    # found pose is partially outside the map
 
-bool                       valid             # whether we found a valid pose
 uint8                      state             # found pose's state: FREE, INSCRIBED, LETHAL, UNKNOWN or OUTSIDE
 uint32                     cost              # found pose's cost (added cost of all cells covered by the footprint)
 geometry_msgs/PoseStamped  pose              # the pose found (filled only if valid is true)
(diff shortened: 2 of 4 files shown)


mbf_msgs/srv/FindValidPose.srv
```diff
@@ -0,0 +1,27 @@
...
+uint8                      costmap           # costmap in which to check the pose
+bool                       current_pose      # check current robot pose instead (ignores pose field)
+bool                       use_padded_fp     # include footprint padding when checking cost; note that safety distance
+                                             # will be measured from the padded footprint
+---
+uint8                      FREE      =  0    # found pose is completely in traversable space
+uint8                      INSCRIBED =  1    # found pose is partially in inscribed space
+uint8                      LETHAL    =  2    # found pose is partially in collision
+uint8                      UNKNOWN   =  3    # found pose is partially in unknown space
+uint8                      OUTSIDE   =  4    # found pose is partially outside the map
```
ChristofDubs, line 28: Question: isn't it somewhat ambiguous which of the invalid codes to return? I.e. it can be that all 3 are the case: There could be a pose within tolerances where parts of the footprint are in collision, another pose where a part of the footprint is outside the map, and yet another pose where a part of the footprint is in unknown space, but no pose that is only in FREE/INSCRIBED. So which of the LETHAL/UNKNOWN/OUTSIDE do you return now? (I guess we could return the error code at `pose`, but that might also be ambiguous)

mbf_msgs/srv/FindValidPose.srv
```diff
@@ -0,0 +1,27 @@
...
+float32                    safety_dist       # minimum distance allowed to the closest obstacle
+float32                    dist_tolerance    # maximum distance we can deviate from the given pose during the search
+float32                    angle_tolerance   # maximum angle we can rotate the given pose during the search
+uint8                      costmap           # costmap in which to check the pose
+bool                       current_pose      # check current robot pose instead (ignores pose field)
+bool                       use_padded_fp     # include footprint padding when checking cost; note that safety distance
+                                             # will be measured from the padded footprint
+---
+uint8                      FREE      =  0    # found pose is completely in traversable space
+uint8                      INSCRIBED =  1    # found pose is partially in inscribed space
```
ChristofDubs, line 25: Suggestion: I think the FREE / INSCRIBED distinction is only meaningful for costs related to the center point of the robot, but not the footprint? Maybe just return `max_cost` and `average_cost` in the result?

mbf_msgs/srv/FindValidPose.srv
```diff
@@ -0,0 +1,27 @@
+# Find the closest valid pose to the given one, within the given linear and angular tolerances.
+# A pose is considered valid if the robot footprint there is entirely inside the map and not in
+# collision nor unknown space.
```
ChristofDubs, line 4: I think this default definition of `vaild pose` makes sense, but I guess technically, we could let the user decide what they consider a valid pose... Maybe allowing poses partially outside the map is an unneeded edge case; I don't know how commonly unknown space is used. Anyway, could just be a future feature to add a flag to allow poses where the footprint contains cells with unknown cost

ChristofDubs: Maybe the conclusion from all my 3 other comments could be:
- remove the boolean `valid` field in the result
- return `FREE` if a pose completely in free space (including inscribed) is found, else return one partially in UNKNOWN, else return one partially outside the map, else return collision. The user can then decide themselves whether they want to use the pose or not based on the return code?

mbf_msgs/srv/FindValidPose.srv
corot: can be,,,  we just provide the info and the user can decide to ignore it by checking state < LETHAL (most common case)

mbf_msgs/srv/FindValidPose.srv
corot: yes,,, makes sense;  or much simpler
```
uint8                      FREE      =  0    # found pose is completely in traversable space
uint8                      INSCRIBED =  1    # found pose is partially in inscribed space
uint8                      INVALID =  2    # no valid pose found
```
and remove valid flag

RainerKuemmerle:
Why would this be best implemented as a service? It might take a considerable amount of time to run (if tolerance is large?) maybe there are intermediate information to transmit in the future to allow the user to abort the search?

mbf_msgs/srv/FindValidPose.srv
corot: > Why would this be best implemented as a service? It might take a considerable amount of time to run (if tolerance is large?) maybe there are intermediate information to transmit in the future to allow the user to abort the search?

good question;  2 reasons:
- we intend this to be fast, so it can be called continuously to adapt to changing situations. From experience, I expect that this won't take more some tens of milliseconds
- it's an extension of the other services provided at mbf_costmap_nav server, so makes sense to offer the same interface

corot:
> Why would this be best implemented as a service? It might take a considerable amount of time to run (if tolerance is large?) maybe there are intermediate information to transmit in the future to allow the user to abort the search?

sorry, I answered on the wrong place 3 weeks ago!!! :see_no_evil: 

good question;  2 reasons:
- we intend this to be fast, so it can be called continuously to adapt to changing situations. From experience, I expect that this won't take more some tens of milliseconds
- it's an extension of the other services provided at mbf_costmap_nav server, so makes sense to offer the same interface

mbf_msgs/srv/FindValidPose.srv
```diff
@@ -0,0 +1,27 @@
...
+                                             # will be measured from the padded footprint
+---
+uint8                      FREE      =  0    # found pose is completely in traversable space
+uint8                      INSCRIBED =  1    # found pose is partially in inscribed space
+uint8                      LETHAL    =  2    # found pose is partially in collision
+uint8                      UNKNOWN   =  3    # found pose is partially in unknown space
+uint8                      OUTSIDE   =  4    # found pose is partially outside the map
+
+bool                       valid             # whether we found a valid pose
+uint8                      state             # found pose's state: FREE, INFLATED, LETHAL, UNKNOWN or OUTSIDE
```
corot:
```suggestion
uint8                      state             # found pose's state: FREE, INSCRIBED, LETHAL, UNKNOWN or OUTSIDE
```

same in the other services
hope comments don't change the checksum!!!

mbf_msgs/srv/FindValidPose.srv
```diff
@@ -0,0 +1,27 @@
...
+bool                       use_padded_fp     # include footprint padding when checking cost; note that safety distance
+                                             # will be measured from the padded footprint
+---
+uint8                      FREE      =  0    # found pose is completely in traversable space
+uint8                      INSCRIBED =  1    # found pose is partially in inscribed space
+uint8                      LETHAL    =  2    # found pose is partially in collision
+uint8                      UNKNOWN   =  3    # found pose is partially in unknown space
+uint8                      OUTSIDE   =  4    # found pose is partially outside the map
+
+bool                       valid             # whether we found a valid pose
```
corot: If we implement caching first UNKNOWN and first OUTSIDE, we can remove this
Then we will return
- FREE or INFLATED as surely valid poses
- UNKNOWN or OUTSIDE as valid or not tu be decided by the client
- LETHAL when there's no better option
```suggestion
```

mbf_msgs/srv/CheckPath.srv
```diff
@@ -21,6 +21,6 @@ uint8                      UNKNOWN   =  3    # path is partially in unknown spac
 uint8                      OUTSIDE   =  4    # path is partially outside the map
 
 uint32                     last_checked      # index of the first pose along the path with return_on state or worse
-uint8                      state             # path worst state (until last_checked): FREE, INFLATED, LETHAL, UNKNOWN...
+uint8                      state             # path worst state (until last_checked): FREE, INSCRIBED, LETHAL, UNKNOWN...
```
corot, line 24: OOS: make comment match the constant name
This should not change checksum: http://wiki.ros.org/ROS/Technical%20Overview#Message_serialization_and_msg_MD5_sums

(2 more)

mbf_msgs/srv/FindValidPose.srv
```diff
@@ -0,0 +1,32 @@
...
+uint8                      costmap           # costmap in which to check the pose
+bool                       current_pose      # check current robot pose instead (ignores pose field)
+bool                       use_padded_fp     # include footprint padding when checking cost; note that safety distance
+                                             # will be measured from the padded footprint
+---
+uint8                      FREE      =  0    # found pose is completely in traversable space
+uint8                      INSCRIBED =  1    # found pose is partially in inscribed space
+uint8                      LETHAL    =  2    # found pose is partially in collision
+uint8                      UNKNOWN   =  3    # found pose is partially in unknown space
+uint8                      OUTSIDE   =  4    # found pose is partially outside the map
```
ChristofDubs, line 28: I'd sort these error codes by precedence, i.e. `FREE=0`, `INSCRIBED=1`, `UNKNOWN=2`, `OUTSIDE=3`, `LETHAL=4` (+same ordering in the comments). Then the user can just do a `<=` check with whatever scenario is still acceptable for them.

corot: yes.... I know would be better,,, but the other services already have this ordering; so I wanted to avoid confusion giving different values to the same constant 
we can change all services, thought. Hopefully nobody is using the numeric value!!!

RainerKuemmerle:
@corot Do you have a benchmark how long the service call will take with the search implementation of #317 ?
Otherwise looks good to me.

RainerKuemmerle (approved):

corot:
> @corot Do you have a benchmark how long the service call will take with the search implementation of #317 ? Otherwise looks good to me.

Yes, it's very fast; we call it at high frequency w/o any problem

mbf_msgs/srv/FindValidPose.srv
```diff
@@ -0,0 +1,32 @@
+# Find the closest free pose to the given one, within the given linear and angular tolerances.
+#
+# A pose is considered free if the robot footprint there is entirely inside the map and not in
```
ChristofDubs:
```suggestion
# A pose is considered free if the robot footprint there is entirely inside the map and neither in
```

mbf_msgs/srv/FindValidPose.srv
```diff
@@ -0,0 +1,32 @@
...
+# collision nor unknown space.
+#
+# If no free pose can be found, but we find one partially on unknown space, or outside the map,
```
ChristofDubs:
```suggestion
# If no free pose can be found, but we find one partially in unknown space, or partially outside the map,
```
or
```suggestion
# If no free pose can be found, but we find one partially in unknown space or outside the map,
```

mbf_msgs/srv/FindValidPose.srv
```diff
@@ -0,0 +1,32 @@
...
+# we will return it and set state to the correspondent option (unknown space takes precedence).
```
ChristofDubs:
```suggestion
# we will return it and set state to the corresponding option (unknown space takes precedence).
```

mbf_msgs/srv/FindValidPose.srv
```diff
@@ -0,0 +1,32 @@
...
+                                             # will be measured from the padded footprint
+---
+uint8                      FREE      =  0    # found pose is completely in traversable space
+uint8                      INSCRIBED =  1    # found pose is partially in inscribed space
+uint8                      LETHAL    =  2    # found pose is partially in collision
+uint8                      UNKNOWN   =  3    # found pose is partially in unknown space
+uint8                      OUTSIDE   =  4    # found pose is partially outside the map
+
+uint8                      state             # found pose's state: FREE, INSCRIBED, LETHAL, UNKNOWN or OUTSIDE
+uint32                     cost              # found pose's cost (added cost of all cells covered by the footprint)
```
ChristofDubs:
```suggestion
uint32                     cost              # found pose's cost (sum of costs over all cells covered by the footprint)
```

PR #317, opened 2023-03-22
Add search helper for goal angle tolerance

renan028:
# Description

Adding a class to search the space around the input goal and return the closest pose free of obstacles that respect both linear and angular tolerances. 
During exploration, eight neighboring cells are added in a priority queue. By default, the cells are ordered by euclidean distance wrt to the start pose, but one can add a different strategy in the constructor.
The search stops when it finds a cell where the robot can be placed (footprint + padding + safety distance) without colliding with obstacles.
For each cell, we test multiple angles.

# Example

In the figure below, the navigation goal is A (the blue dot) and the linear tolerance is the circle. The orientation of the goal is the Y-axis. 
The red polygons are static obstacles (e.g. walls), orange polygons are dynamic obstacles (e.g. chairs, tables, boxes) and gray polygons are unknown space. The problem is to find the closest free space to the goal that the robot can be, i.e., its footprint is not in collision with any obstacle and the distance is equal or above a given clearance distance.

image

The goal orientation is not a valid solution:

image

But we can find a solution using the angle tolerance:
image

### DIFF
--- a/mbf_costmap_nav/CMakeLists.txt
+++ b/mbf_costmap_nav/CMakeLists.txt
@@ -79,6 +79,7 @@ add_library(${MBF_COSTMAP_2D_SERVER_LIB}
   src/mbf_costmap_nav/costmap_wrapper.cpp
   src/mbf_costmap_nav/footprint_helper.cpp
   src/mbf_costmap_nav/search_helper.cpp
+  src/mbf_costmap_nav/search_helper_viz.cpp
 )
 add_dependencies(${MBF_COSTMAP_2D_SERVER_LIB} ${catkin_EXPORTED_TARGETS})
 add_dependencies(${MBF_COSTMAP_2D_SERVER_LIB} ${MBF_NAV_CORE_WRAPPER_LIB})
--- a/mbf_costmap_nav/CMakeLists.txt
+++ b/mbf_costmap_nav/CMakeLists.txt
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 3.0.2)
 project(mbf_costmap_nav)
 
+set(CMAKE_CXX_STANDARD 17)
+
 find_package(catkin REQUIRED
   COMPONENTS
   costmap_2d
@@ -76,6 +78,7 @@ add_library(${MBF_COSTMAP_2D_SERVER_LIB}
   src/mbf_costmap_nav/costmap_recovery_execution.cpp
   src/mbf_costmap_nav/costmap_wrapper.cpp
   src/mbf_costmap_nav/footprint_helper.cpp
+  src/mbf_costmap_nav/search_helper.cpp
 )
 add_dependencies(${MBF_COSTMAP_2D_SERVER_LIB} ${catkin_EXPORTED_TARGETS})
 add_dependencies(${MBF_COSTMAP_2D_SERVER_LIB} ${MBF_NAV_CORE_WRAPPER_LIB})
@@ -108,3 +111,19 @@ install(DIRECTORY include/${PROJECT_NAME}/
 catkin_install_python(PROGRAMS scripts/move_base_legacy_relay.py
   DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
 )
+
+#############
+## Test    ##
+#############
+
+if(CATKIN_ENABLE_TESTING)
+  find_package(rostest REQUIRED)
+
+  add_rostest_gtest(search_helper_test
+    test/search_helper.test
+    test/search_helper_test.cpp
+  )
+  target_link_libraries(search_helper_test
+    ${MBF_COSTMAP_2D_SERVER_LIB}
+  )
+endif()
--- a/mbf_costmap_nav/CMakeLists.txt
+++ b/mbf_costmap_nav/CMakeLists.txt
@@ -78,8 +78,8 @@ add_library(${MBF_COSTMAP_2D_SERVER_LIB}
   src/mbf_costmap_nav/costmap_recovery_execution.cpp
   src/mbf_costmap_nav/costmap_wrapper.cpp
   src/mbf_costmap_nav/footprint_helper.cpp
-  src/mbf_costmap_nav/search_helper.cpp
-  src/mbf_costmap_nav/search_helper_viz.cpp
+  src/mbf_costmap_nav/free_pose_search.cpp
+  src/mbf_costmap_nav/free_pose_search_viz.cpp
 )
 add_dependencies(${MBF_COSTMAP_2D_SERVER_LIB} ${catkin_EXPORTED_TARGETS})
 add_dependencies(${MBF_COSTMAP_2D_SERVER_LIB} ${MBF_NAV_CORE_WRAPPER_LIB})
@@ -120,11 +120,11 @@ catkin_install_python(PROGRAMS scripts/move_base_legacy_relay.py
 if(CATKIN_ENABLE_TESTING)
   find_package(rostest REQUIRED)
 
-  add_rostest_gtest(search_helper_test
-    test/search_helper.test
-    test/search_helper_test.cpp
+  add_rostest_gtest(free_pose_search_test
+    test/free_pose_search.test
+    test/free_pose_search_test.cpp
   )
-  target_link_libraries(search_helper_test
+  target_link_libraries(free_pose_search_test
     ${MBF_COSTMAP_2D_SERVER_LIB}
   )
 endif()
--- a/mbf_costmap_nav/CMakeLists.txt
+++ b/mbf_costmap_nav/CMakeLists.txt
@@ -119,6 +119,7 @@ catkin_install_python(PROGRAMS scripts/move_base_legacy_relay.py
 
 if(CATKIN_ENABLE_TESTING)
   find_package(rostest REQUIRED)
+  find_package(map_server REQUIRED)
 
   add_rostest_gtest(free_pose_search_test
     test/free_pose_search.test
--- a/mbf_costmap_nav/include/mbf_costmap_nav/costmap_navigation_server.h
+++ b/mbf_costmap_nav/include/mbf_costmap_nav/costmap_navigation_server.h
@@ -47,6 +47,7 @@
 #include <mbf_msgs/CheckPath.h>
 #include <mbf_msgs/CheckPose.h>
 #include <mbf_msgs/CheckPoint.h>
+#include <mbf_msgs/FindValidPose.h>
 
 #include <nav_core/base_global_planner.h>
 #include <nav_core/base_local_planner.h>
@@ -192,6 +193,15 @@ private:
       const mbf_abstract_core::AbstractRecovery::Ptr &behavior_ptr);
 
   /**
+   * @brief If mbf_msgs::CheckPose::Request::LOCAL_COSTMAP the local costmap is returned
+   * if mbf_msgs::CheckPose::Request::GLOBAL_COSTMAP, the global costmap is returned.
+   * Otherwise, it returns an empty pointer.
+   * @param costmap_type The type of the costmap to return
+   * @return A shared pointer to the requested costmap, if the costmap_type is valid, an empty pointer otherwise.
+  */
+  CostmapWrapper::Ptr selectingRequestedCostmap(mbf_msgs::CheckPose::Request::_costmap_type costmap_type) const;
+
+  /**
    * @brief Callback method for the check_point_cost service
    * @param request Request object, see the mbf_msgs/CheckPoint service definition file.
    * @param response Response object, see the mbf_msgs/CheckPoint service definition file.
@@ -225,6 +235,15 @@ private:
    * @return true, if the service completed successfully, false otherwise
    */
   bool callServiceClearCostmaps(std_srvs::Empty::Request &request, std_srvs::Empty::Response &response);
+
+  /**
+   * @brief Callback method for the find valid pose service
+   * @param request FindValidPose request object.
+   * @param response FindValidPose response object.
+   * @return true, if the service completed successfully, false otherwise
+   */
+  bool callServiceFindValidPose(mbf_msgs::FindValidPose::Request &request, 
+                                mbf_msgs::FindValidPose::Response &response);
 
   /**
    * @brief Reconfiguration method called by dynamic reconfigure.
@@ -275,6 +294,11 @@ private:
 
   //! Service Server for the clear_costmap service
   ros::ServiceServer clear_costmaps_srv_;
+
+  //! Service Server for the find_valid_pose service
+  ros::ServiceServer find_valid_pose_srv_;
+
+  static constexpr double ANGLE_INCREMENT = 5.0 * M_PI / 180.0; // 5 degrees
 };
 
 } /* namespace mbf_costmap_nav */
--- a/mbf_costmap_nav/include/mbf_costmap_nav/costmap_navigation_server.h
+++ b/mbf_costmap_nav/include/mbf_costmap_nav/costmap_navigation_server.h
@@ -72,8 +72,8 @@ namespace mbf_costmap_nav
  * @brief Classes belonging to the Move Base Server level.
  */
 
-
-typedef boost::shared_ptr<dynamic_reconfigure::Server<mbf_costmap_nav::MoveBaseFlexConfig> > DynamicReconfigureServerCostmapNav;
+typedef boost::shared_ptr<dynamic_reconfigure::Server<mbf_costmap_nav::MoveBaseFlexConfig> >
+    DynamicReconfigureServerCostmapNav;
 
 /// @brief A mapping from a string to a map-ptr.
 typedef boost::unordered_map<std::string, CostmapWrapper::Ptr> StringToMap;
@@ -89,14 +89,13 @@ typedef boost::unordered_map<std::string, CostmapWrapper::Ptr> StringToMap;
 class CostmapNavigationServer : public mbf_abstract_nav::AbstractNavigationServer
 {
 public:
-
   typedef boost::shared_ptr<CostmapNavigationServer> Ptr;
 
   /**
    * @brief Constructor
    * @param tf_listener_ptr Shared pointer to a common TransformListener
    */
-  CostmapNavigationServer(const TFPtr &tf_listener_ptr);
+  CostmapNavigationServer(const TFPtr& tf_listener_ptr);
 
   /**
    * @brief Destructor
@@ -106,16 +105,14 @@ public:
   virtual void stop();
 
 private:
-
   /**
    * @brief Create a new planner execution.
    * @param plugin_name Name of the planner to use.
    * @param plugin_ptr Shared pointer to the plugin to use.
    * @return Shared pointer to a new @ref planner_execution "PlannerExecution".
    */
-  virtual mbf_abstract_nav::AbstractPlannerExecution::Ptr newPlannerExecution(
-      const std::string &plugin_name,
-      const mbf_abstract_core::AbstractPlanner::Ptr &plugin_ptr);
+  virtual mbf_abstract_nav::AbstractPlannerExecution::Ptr
+  newPlannerExecution(const std::string& plugin_name, const mbf_abstract_core::AbstractPlanner::Ptr& plugin_ptr);
 
   /**
    * @brief Create a new controller execution.
@@ -123,9 +120,8 @@ private:
    * @param plugin_ptr Shared pointer to the plugin to use.
    * @return Shared pointer to a new @ref controller_execution "ControllerExecution".
    */
-  virtual mbf_abstract_nav::AbstractControllerExecution::Ptr newControllerExecution(
-      const std::string &plugin_name,
-      const mbf_abstract_core::AbstractController::Ptr &plugin_ptr);
+  virtual mbf_abstract_nav::AbstractControllerExecution::Ptr
+  newControllerExecution(const std::string& plugin_name, const mbf_abstract_core::AbstractController::Ptr& plugin_ptr);
 
   /**
    * @brief Create a new recovery behavior execution.
@@ -133,16 +129,15 @@ private:
    * @param plugin_ptr Shared pointer to the plugin to use
    * @return Shared pointer to a new @ref recovery_execution "RecoveryExecution".
    */
-  virtual mbf_abstract_nav::AbstractRecoveryExecution::Ptr newRecoveryExecution(
-      const std::string &plugin_name,
-      const mbf_abstract_core::AbstractRecovery::Ptr &plugin_ptr);
+  virtual mbf_abstract_nav::AbstractRecoveryExecution::Ptr
+  newRecoveryExecution(const std::string& plugin_name, const mbf_abstract_core::AbstractRecovery::Ptr& plugin_ptr);
 
   /**
    * @brief Loads the plugin associated with the given planner_type parameter.
    * @param planner_type The type of the planner plugin to load.
    * @return true, if the local planner plugin was successfully loaded.
    */
-  virtual mbf_abstract_core::AbstractPlanner::Ptr loadPlannerPlugin(const std::string &planner_type);
+  virtual mbf_abstract_core::AbstractPlanner::Ptr loadPlannerPlugin(const std::string& planner_type);
 
   /**
    * @brief Initializes the controller plugin with its name and pointer to the costmap
@@ -150,10 +145,8 @@ private:
    * @param planner_ptr pointer to the planner object which corresponds to the name param
    * @return true if init succeeded, false otherwise
    */
-  virtual bool initializePlannerPlugin(
-      const std::string &name,
-      const mbf_abstract_core::AbstractPlanner::Ptr &planner_ptr
-  );
+  virtual bool initializePlannerPlugin(const std::string& name,
+                                       const mbf_abstract_core::AbstractPlanner::Ptr& planner_ptr);
 
   /**
    * @brief Loads the plugin associated with the given controller type parameter
@@ -161,7 +154,7 @@ private:
    * @return A shared pointer to a new loaded controller, if the controller plugin was loaded successfully,
    *         an empty pointer otherwise.
    */
-  virtual mbf_abstract_core::AbstractController::Ptr loadControllerPlugin(const std::string &controller_type);
+  virtual mbf_abstract_core::AbstractController::Ptr loadControllerPlugin(const std::string& controller_type);
 
   /**
    * @brief Initializes the controller plugin with its name, a pointer to the TransformListener
@@ -170,17 +163,15 @@ private:
    * @param controller_ptr pointer to the controller object which corresponds to the name param
    * @return true if init succeeded, false otherwise
    */
-  virtual bool initializeControllerPlugin(
-      const std::string &name,
-      const mbf_abstract_core::AbstractController::Ptr &controller_ptr
-  );
+  virtual bool initializeControllerPlugin(const std::string& name,
+                                          const mbf_abstract_core::AbstractController::Ptr& controller_ptr);
 
   /**
    * @brief Loads a Recovery plugin associated with given recovery type parameter
    * @param recovery_name The name of the Recovery plugin
    * @return A shared pointer to a Recovery plugin, if the plugin was loaded successfully, an empty pointer otherwise.
    */
-  virtual mbf_abstract_core::AbstractRecovery::Ptr loadRecoveryPlugin(const std::string &recovery_type);
+  virtual mbf_abstract_core::AbstractRecovery::Ptr loadRecoveryPlugin(const std::string& recovery_type);
 
   /**
    * @brief Initializes a recovery behavior plugin with its name and pointers to the global and local costmaps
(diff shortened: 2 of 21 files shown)


mbf_costmap_nav/include/mbf_costmap_nav/search_helper.h
```diff
@@ -0,0 +1,191 @@
...
+  double linear_tolerance;
+  bool use_padded_fp;
+  double safety_dist;
+  geometry_msgs::Pose2D goal;
+};
+
+struct SearchState
+{
+  unsigned char cost;
+  std::uint8_t state{ State::UNSET };
```
RainerKuemmerle: Missing cstdint include in this file?

mbf_costmap_nav/include/mbf_costmap_nav/search_helper.h
```diff
@@ -0,0 +1,191 @@
...
+
+class EuclideanCompare
+{
+private:
+  Cell start_;
+  int start_x_;
+  int start_y_;
+
+public:
+  EuclideanCompare(const Cell& start) : start_(start)
```
RainerKuemmerle: explicit

mbf_costmap_nav/include/mbf_costmap_nav/search_helper.h
```diff
@@ -0,0 +1,191 @@
...
+  };
+};
+
+struct SearchSolution
+{
+  geometry_msgs::Pose2D pose;
+  SearchState search_state;
+};
+
+class EuclideanCompare
```
RainerKuemmerle: docstring. What is this class doing?

mbf_costmap_nav/include/mbf_costmap_nav/search_helper.h
```diff
@@ -0,0 +1,191 @@
...
+#include <geometry_msgs/Point.h>
+#include <geometry_msgs/Pose2D.h>
+#include <costmap_2d/costmap_2d_ros.h>
+#include <std_msgs/ColorRGBA.h>
+#include <visualization_msgs/MarkerArray.h>
+
+namespace mbf_costmap_nav
+{
+
+struct SearchConfig
```
RainerKuemmerle: Default initialize values?

mbf_costmap_nav/include/mbf_costmap_nav/search_helper.h
```diff
@@ -0,0 +1,191 @@
...
+ * can add a new strategy).
+ * The search stops when it finds a cell where the robot can be placed without colliding with obstacles (footprint +
+ * padding).
+ * For each cell, we test multiple angles.
+ * The search is performed on the costmap given in the constructor.
+ */
+class SearchHelper
+{
+private:
+  static constexpr auto LOGNAME = "search_helper";
```
RainerKuemmerle: auto will resolve to some char array type? Can we use std::string_view - was that maybe intended?

mbf_costmap_nav/include/mbf_costmap_nav/search_helper.h
```diff
@@ -0,0 +1,191 @@
...
+ * robot without colliding with obstacles.
+ * It is a Breadth-depth search performed in a spiral pattern (neighboring cells added in a priority queue), starting
+ * from the start (given) cell. And the cells are ordered by euclidean distance from the start cell (by default but one
+ * can add a new strategy).
+ * The search stops when it finds a cell where the robot can be placed without colliding with obstacles (footprint +
+ * padding).
+ * For each cell, we test multiple angles.
+ * The search is performed on the costmap given in the constructor.
+ */
+class SearchHelper
```
RainerKuemmerle: FreePoseSearch ?
Calling everything a helper seems not to help a lot in understanding what the class is doing.

RainerKuemmerle: Can the members be protected? Making the search method virtual? Maybe later there are other implementations?

RainerKuemmerle: Why the mix of costmap_2d::Costmap2D* and costmap_2d::Costmap2DROS* in the interface of this class?

mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
```diff
@@ -825,4 +816,82 @@ bool CostmapNavigationServer::callServiceClearCostmaps(std_srvs::Empty::Request
...
+      break;
+    default:
+      ROS_ERROR_STREAM("No valid costmap provided; options are "
+                       << mbf_msgs::CheckPose::Request::LOCAL_COSTMAP << ": local costmap, "
+                       << mbf_msgs::CheckPose::Request::GLOBAL_COSTMAP << ": global costmap");
+      return costmap;
+  }
+}
+
+bool CostmapNavigationServer::callServiceFindValidPose(mbf_msgs::FindValidPose::Request& request,
```
RainerKuemmerle, line 839: can request be const?

mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
```diff
@@ -825,4 +816,82 @@ bool CostmapNavigationServer::callServiceClearCostmaps(std_srvs::Empty::Request
...
+                                                       mbf_msgs::FindValidPose::Response& response)
+{
+  CostmapWrapper::Ptr costmap = selectingRequestedCostmap(request.costmap);
+  if (!costmap)
+  {
+    return false;
+  }
+  std::string costmap_name =
```
RainerKuemmerle: can be declared const. Not sure, if you want to do that good practice in general.

mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
```diff
@@ -825,4 +816,82 @@ bool CostmapNavigationServer::callServiceClearCostmaps(std_srvs::Empty::Request
...
+                       request.dist_tolerance, static_cast<bool>(request.use_padded_fp),
+                       request.safety_dist,    goal };
+  SearchHelperViz viz(private_nh, costmap_frame);
+  SearchHelper search_helper(costmap.get(), config, std::nullopt, viz);
+
+  // search for a valid pose
+  const auto sol = search_helper.search();
+
+  response.pose.pose.position.x = sol.pose.x;
+  response.pose.pose.position.y = sol.pose.y;
```
RainerKuemmerle, line 886: z can be arbitrary? Set to zero explicitly.

mbf_costmap_nav/src/mbf_costmap_nav/footprint_helper.cpp
```diff
@@ -111,6 +111,102 @@ void FootprintHelper::getLineCells(int x0, int x1, int y0, int y1, std::vector<C
   }
 }
 
+void FootprintHelper::supercover(int x0, int x1, int y0, int y1, std::vector<Cell>& pts)
+{
+  {
+    auto POINT = [&pts](int y, int x) {
```
RainerKuemmerle: bad style. MACRO vs function in c++

mbf_costmap_nav/src/mbf_costmap_nav/search_helper.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+// std
+#include <queue>
+#include <unordered_set>
+
+namespace mbf_costmap_nav
+{
+
+SearchHelper::SearchHelper(const costmap_2d::Costmap2DROS* costmap, const SearchConfig& config,
+                           const std::optional<std::function<bool(const Cell, const Cell)>>& compare_strategy,
```
RainerKuemmerle: Why are the function arguments here passed by value?

mbf_costmap_nav/src/mbf_costmap_nav/search_helper.cpp
```diff
@@ -0,0 +1,294 @@
...
+#include "mbf_costmap_nav/search_helper.h"
+
+// std
+#include <queue>
+#include <unordered_set>
+
+namespace mbf_costmap_nav
+{
+
+SearchHelper::SearchHelper(const costmap_2d::Costmap2DROS* costmap, const SearchConfig& config,
```
RainerKuemmerle: Can this class work on a reference instead? Why passing a RAW pointer all the time to all the functions? You never check for nullptr at any place either.

mbf_costmap_nav/src/mbf_costmap_nav/search_helper.cpp
```diff
@@ -0,0 +1,294 @@
...
+    }
+  }
+  return neighbors;
+}
+
+std::vector<geometry_msgs::Point> SearchHelper::safetyPadding(const costmap_2d::Costmap2DROS* costmap_2dros,
+                                                              const bool use_padded_fp, const double safety_dist)
+
+{
+  std::vector<geometry_msgs::Point> footprint;
```
RainerKuemmerle: merge with line below

mbf_costmap_nav/src/mbf_costmap_nav/search_helper.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+  // create a map of cells to avoid duplicates
+  std::unordered_map<int, Cell> cells_to_check_map;
+  for (const auto& cell : cells_to_check)
+  {
+    cells_to_check_map[costmap_2d->getIndex(cell.x, cell.y)] = cell;
+  }
+
+  unsigned char max_cost = 0;
+  for (const auto& [index, cell] : cells_to_check_map)
```
RainerKuemmerle: You are not using index? Why is cells_to_check_map an unordered_map and not just a vector<Cell>

mbf_costmap_nav/src/mbf_costmap_nav/search_helper.cpp
```diff
@@ -0,0 +1,294 @@
...
+SearchSolution SearchHelper::findValidOrientation(const costmap_2d::Costmap2D* costmap_2d,
+                                                  const std::vector<geometry_msgs::Point>& footprint,
+                                                  const geometry_msgs::Pose2D& pose_2d, const SearchConfig& config,
+                                                  std::optional<SearchHelperViz>& viz)
+{
+  bool found_unknown{ false };
+  SearchSolution sol;
+  sol.pose = pose_2d;
+
+  double dyaw = 0;
```
RainerKuemmerle: declaration in the for statement. Not used outside.

mbf_costmap_nav/src/mbf_costmap_nav/search_helper.cpp
```diff
@@ -0,0 +1,294 @@
...
+  // lock costmap so content doesn't change while adding cell costs
+  boost::unique_lock<costmap_2d::Costmap2D::mutex_t> lock(*(costmap2d->getMutex()));
+
+  std::unordered_set<int> in_queue_or_visited;
+
+  std::vector<geometry_msgs::Point> footprint = safetyPadding(costmap_, config_.use_padded_fp, config_.safety_dist);
+
+  SearchSolution sol;
+  sol.pose.theta = config_.goal.theta;
+  bool found_unknown{ false };
```
RainerKuemmerle: only place this syntax is used instead of bool found_unknown = false.

RainerKuemmerle (commented):
Cool feature! I just had time to quickly look at some syntax suggestions.

mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
```diff
@@ -825,4 +816,82 @@ bool CostmapNavigationServer::callServiceClearCostmaps(std_srvs::Empty::Request
...
+  response.pose.pose.orientation = tf::createQuaternionMsgFromYaw(sol.pose.theta);
```
corot, line 888: This returns consistently 0 yaw as orientation:
in
```
header: 
  seq: 0
  stamp: 0.000000000
  frame_id: map
pose: 
  position: 
    x: 2.9889
    y: 9.97502
    z: 0
  orientation: 
    x: 0
    y: 0
    z: 0.975985
    w: 0.217839
```
out
```
header: 
  seq: 0
  stamp: 1681505298.031356999
  frame_id: map
pose: 
  position: 
    x: 2.725
    y: 9.725
    z: 0
  orientation: 
    x: 0
    y: 0
    z: 0
    w: 1
```

and a video

https://user-images.githubusercontent.com/322610/232153951-a353d56d-5e27-4ab7-abf0-56f29c5b7f0c.mp4

mbf_costmap_nav/include/mbf_costmap_nav/search_helper.h
renan028: I thought it would resolve to `string_view`, but it resolves to `char* const`
I am changing to string_view

mbf_costmap_nav/include/mbf_costmap_nav/search_helper.h
renan028: > Why the mix of costmap_2d::Costmap2D* and costmap_2d::Costmap2DROS* in the interface of this class?

Since the methods are static, the idea is to require the minimum argument for the function to be able to run. 
We have methods that require the `costmap2DROS` wrapper, which is an extension with some extra functionalities; but also methods that can work with the pure `costmap2D` (ROS independent), which is a lighter version.

If they were not static, I would store `costmap2DROS` and use it in all methods, but I prefer to make them static since they may be useful outside; and they are independent functionalities

mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
renan028: no, the callback of ros service server cannot be const

mbf_costmap_nav/src/mbf_costmap_nav/footprint_helper.cpp
renan028: I didn't want to change the copy-pasted code for reference, but I will fix the style at least

mbf_costmap_nav/src/mbf_costmap_nav/search_helper.cpp
renan028: I don't think there is an improvement in this case because it is only 12 bytes. But it can become bigger in the future, so I will change.

mbf_costmap_nav/src/mbf_costmap_nav/search_helper.cpp
renan028: changing to const ref

mbf_costmap_nav/src/mbf_costmap_nav/search_helper.cpp
renan028: The reason is in the comment above. I am creating an `unordered_map` to avoid duplicates. Best would be an `unordered_set`, but that requires a hash given by the `costmap` (index), so the `unordered_map` solves the problem here.
I am changing `[index, cell]` to `[_, cell]`, though

mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
renan028: I have added a test for that:
https://github.com/magazino/move_base_flex/pull/317/files#diff-8fb727cf2dfe977e499637719122f6e082cda1178d701aa02d0c0d45a48d2204R393

The unittest works. Maybe I am missing something. Can you share how can I reproduce the issue?

corot: standalone navigation + clicked point nav goals, as in the video; nothing special 

Or just calling the service with any non-zero rotation returns a 0-yaw pose:
```
$ rosservice call /move_base_flex/find_valid_pose "pose:
  header:
    seq: 0
    stamp: {secs: 0, nsecs: 0}
    frame_id: 'map'
  pose:
    position: {x: 3.0, y: 6.0, z: 0.0}
    orientation: {x: 0.0, y: 0.0, z: 0.707, w: 0.707}
safety_dist: 0.0
dist_tolerance: 1.0
angle_tolerance: 0.0
costmap: 1
current_pose: false
use_padded_fp: false" 
```
gives on standalone nav
```
state: 1
cost: 253
pose: 
  header: 
    seq: 0
    stamp: 
      secs: 489
      nsecs: 450000000
    frame_id: "map"
  pose: 
    position: 
      x: 2.9875000000000043
      y: 6.212500000000004
      z: 0.0
    orientation: 
      x: 0.0
      y: 0.0
      z: 0.0
      w: 1.0
```

mbf_costmap_nav/CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 3.0.2)
 project(mbf_costmap_nav)
 
+set(CMAKE_CXX_STANDARD 17)
```
corot, line 4: Q / F: MBF uses [C++03](https://en.wikipedia.org/wiki/C%2B%2B03) standard.

I see we use lambdas here, so we need at least C++11.

I guess for master is just fine to use C++17. @spuetz, @RainerKuemmerle, thoughts on this?
And can we just set the standard on the metapackage?

mbf_costmap_nav/include/mbf_costmap_nav/costmap_navigation_server.h
```diff
@@ -191,6 +192,15 @@ class CostmapNavigationServer : public mbf_abstract_nav::AbstractNavigationServe
...
       const mbf_abstract_core::AbstractRecovery::Ptr &behavior_ptr);
 
+  /**
+   * @brief If mbf_msgs::CheckPose::Request::LOCAL_COSTMAP the local costmap is returned
+   * if mbf_msgs::CheckPose::Request::GLOBAL_COSTMAP, the global costmap is returned.
+   * Otherwise, it returns an empty pointer.
+   * @param costmap_type The type of the costmap to return
+   * @return A shared pointer to the requested costmap, if the costmap_type is valid, an empty pointer otherwise.
+  */
+  CostmapWrapper::Ptr selectingRequestedCostmap(mbf_msgs::CheckPose::Request::_costmap_type costmap_type) const;
```
corot: strong S: we are not selecting anything; just getting (can also be `getRequestedCostmap`, but `requestedCostmap` gives the same info and is more compact)
```suggestion
  CostmapWrapper::Ptr requestedCostmap(mbf_msgs::CheckPose::Request::_costmap_type costmap_type) const;
```

mbf_costmap_nav/include/mbf_costmap_nav/free_pose_search.h
```diff
@@ -0,0 +1,199 @@
+/*********************************************************************
+ *
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2023, Willow Garage, Inc.
```
corot: F: not sure what the copyright should be, but for sure not Willow Garage!!!
@spuetz , @RainerKuemmerle , what to copyright should set here?
Rapyuta Robotics Co., Ltd.?
Or @renan028's name?

Also we should follow the format used on original MBF files, not the one used on copied files, as `footprint_helper.h`

mbf_costmap_nav/include/mbf_costmap_nav/footprint_helper.h
```diff
@@ -66,6 +66,17 @@ class FootprintHelper
...
+    /**
+   * @brief  Supercover algorithm is a modified Bresenham which prints ALL the points (not only one point per axis) the ideal line contains
+   * ref: http://eugen.dedu.free.fr/projects/bresenham/
+   * @param  x0 The x coordinate of the first point
+   * @param  x1 The x coordinate of the second point
+   * @param  y0 The y coordinate of the first point
+   * @param  y1 The y coordinate of the second point
+   * @param  pts Will be filled with the cells that lie on the line in the grid
+   */
+  static void supercover(int x0, int x1, int y0, int y1, std::vector<Cell>& pts);
```
corot, line 78: Q / F: maybe we should just replace `getLineCells`. Otherwise we will use different coverage algorithms on different services, thus giving different costs

@spuetz , @RainerKuemmerle , thoughts?

mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
```diff
@@ -673,28 +673,16 @@ bool CostmapNavigationServer::callServiceCheckPoseCost(mbf_msgs::CheckPose::Requ
...
-      costmap_name = "global costmap";
-      break;
-    default:
-      ROS_ERROR_STREAM("No valid costmap provided; options are "
-                       << mbf_msgs::CheckPath::Request::LOCAL_COSTMAP << ": local costmap, "
-                       << mbf_msgs::CheckPath::Request::GLOBAL_COSTMAP << ": global costmap");
-      return false;
+    return false;
   }
+  std::string costmap_name =
```
corot: S: Assuming we allow C++17 std, `selectingRequestedCostmap` could return a std::pair<std::string, CostmapWrapper::Ptr> and unpack it using structured binding

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
```diff
@@ -0,0 +1,296 @@
...
+      switch (search_state.state)
+      {
+        case SearchState::FREE:
+        case SearchState::INSCRIBED:
+          if (viz)
+          {
+            viz->addSolution(sol.pose, footprint);
+          }
+          return { sol.pose, search_state };
+          break;
```
corot: F: remove

mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
corot: Replacing initialization_list [here](https://github.com/magazino/move_base_flex/pull/317/files#diff-43fb1b9fdaf229fde2a9661293a0da41b5bfba67e7843bdb79dd19ce01fdd3cbR105) with vector solves the issue

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
```diff
@@ -0,0 +1,296 @@
...
+  const std::vector<geometry_msgs::Point> footprint =
+      safetyPadding(costmap_, config_.use_padded_fp, config_.safety_dist);
+
+  SearchSolution sol;
+  sol.pose.theta = config_.goal.theta;
+  bool found_unknown{ false };
+
+  // initializing queue with the goal cell
+  Cell start;
+  costmap2d->worldToMap(config_.goal.x, config_.goal.y, start.x, start.y);
```
corot: F: this can cause segfault for goals outside the map

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
```diff
@@ -0,0 +1,296 @@
...
+                                                                        SearchState::FREE;
+  return { max_cost, state };
+}
+
+SearchSolution FreePoseSearch::findValidOrientation(const costmap_2d::Costmap2D& costmap_2d,
+                                                    const std::vector<geometry_msgs::Point>& footprint,
+                                                    const geometry_msgs::Pose2D& pose_2d, const SearchConfig& config,
+                                                    std::optional<FreePoseSearchViz>& viz)
+{
+  bool found_unknown = false;
```
corot: S: `outside_or_unknown`?

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
```diff
@@ -0,0 +1,298 @@
...
+      safetyPadding(costmap_, config_.use_padded_fp, config_.safety_dist);
+
+  SearchSolution sol;
+  sol.pose.theta = config_.goal.theta;
+  bool found_unknown{ false };
+
+  // initializing queue with the goal cell
+  Cell start;
+  if (costmap2d->worldToMap(config_.goal.x, config_.goal.y, start.x, start.y))
+    start.cost = costmap2d->getCost(start.x, start.y);
```
corot, line 233: F: This was making MBF crash when the goal is outside the costmap

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
```diff
@@ -0,0 +1,298 @@
...
+  else
+    start.cost = costmap_2d::FREE_SPACE;  // outside of the map; TODO: if set to NO_INFORMATION, we won't find a solution
```
corot: S: would be nicer with NO_INFORMATION, but then we need to change the search code

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
```diff
@@ -0,0 +1,298 @@
...
+                               "The best solution found is outside the map");
+    if (viz_)
+    {
+      viz_->addSolution(sol.pose, footprint);
+      viz_->publish();
+    }
+    return sol;
+  }
+
+  ROS_INFO_STREAM("No solution found within tolerance of goal; ending search");
```
corot: F: this should be DEBUG; it will be printed many times when search fails

mbf_costmap_nav/CMakeLists.txt
renan028: I am also using `std::optional` (c++17)

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
renan028: hum, yes. If the goal is outside the map, the cell does not exist and `getNeighbors` will be empty.
I will fix that.
There was actually an issue when solution was outside. I have fixed it and added test for it

mbf_costmap_nav/include/mbf_costmap_nav/free_pose_search.h
renan028: fixed following the template of others

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
```diff
@@ -0,0 +1,359 @@
...
+    }
+  }
+
+  if (outside_or_unknown)
+  {
+    // the solution is a no information pose or outside
+    ROS_INFO_STREAM_COND_NAMED(sol.search_state.state == SearchState::UNKNOWN, LOGNAME.data(),
+                               "The best solution found has NO_INFORMATION cost");
+    ROS_INFO_STREAM_COND_NAMED(sol.search_state.state == SearchState::OUTSIDE, LOGNAME.data(),
+                               "The best solution found is outside the map");
```
corot: These 2 guys should be DEBUG, as they can print many times in a search

mbf_costmap_nav/include/mbf_costmap_nav/footprint_helper.h
```diff
@@ -66,6 +66,17 @@ class FootprintHelper
                                              const std::vector<geometry_msgs::Point>& footprint_spec,
                                              const costmap_2d::Costmap2D&, bool fill);
 
+    /**
+   * @brief  Supercover algorithm is a modified Bresenham which prints ALL the points (not only one point per axis) the ideal line contains
+   * ref: http://eugen.dedu.free.fr/projects/bresenham/
```
RainerKuemmerle, line 71: License is ok?

mbf_costmap_nav/CMakeLists.txt
RainerKuemmerle: c++17 is ok for me.

mbf_costmap_nav/include/mbf_costmap_nav/footprint_helper.h
RainerKuemmerle: Is this one here needed to make the approach work?

renan028: yes, It is needed. Actually, Bresenham line drawing misses grid squares that the ray passes through. The "supercover" algorithm gets all the squares. Great description in Realtime Collision Detection "7.4.2 Uniform Grid Intersection Test".

image

mbf_costmap_nav/include/mbf_costmap_nav/footprint_helper.h
renan028: I have sent an email to the author and his response:
```
Hi,

The code is free, use GPL if you can, or MIT etc.  I am glad it is
useful, and would be glad if you can credit me.

Best regards,
Eugen
```

renan028: Hi, @RainerKuemmerle can you re-review, please? We would like to integrate this feature in mbf in a couple of days and your comments and approval  are very important for us.
Thanks in advance

mbf_costmap_nav/include/mbf_costmap_nav/footprint_helper.h
renan028: I think we should replace `getLineCells` 
But I am afraid that change would be a PR of "type" `change` instead of just `add` (adding a new feature). It will change the behavior for people using the stack as it is (maybe unwanted somehow).
So I would leave the change to another PR (so that maybe you can make a release on top of that change).

corot: makes sense

corot (approved):

RainerKuemmerle:
@renan028 @corot The tests fail on melodic. Is this something to investigate?

RainerKuemmerle (approved):

corot:
> @renan028 @corot The tests fail on melodic. Is this something to investigate?

On it

btw, @RainerKuemmerle, can you also review the service interface we are implementing here? PR #316
Thanks!

corot:
> @renan028 @corot The tests fail on melodic. Is this something to investigate?

just realized that melodic EOL was 3 days ago!

image

But anyway, we investigated the problem. Looks like the test costmap is not getting initialized with the map published on /map topic, but we didn't find out why yet. In any case, it's clearly not an issue with the new service.

PR #318, opened 2023-04-18
Fix replanning shutdown

renan028:
# Description 
Thread `join` hangs forever if one uses the class as an object (not the ROS node), because `ros::ok()` is always true. 
Example, if one try to use it for tests, the test hangs forever and it fails in timeout.

### DIFF
--- a/mbf_abstract_nav/include/mbf_abstract_nav/move_base_action.h
+++ b/mbf_abstract_nav/include/mbf_abstract_nav/move_base_action.h
@@ -166,6 +166,7 @@ class MoveBaseAction
 
   //! Replanning thread, running permanently
   boost::thread replanning_thread_;
+  bool replanning_thread_shutdown_{ false };
 
   //! true, if recovery behavior for the MoveBase action is enabled.
   bool recovery_enabled_;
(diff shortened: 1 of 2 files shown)


mbf_abstract_nav/include/mbf_abstract_nav/move_base_action.h
```diff
@@ -166,6 +166,7 @@ class MoveBaseAction
 
   //! Replanning thread, running permanently
   boost::thread replanning_thread_;
+  bool replanning_thread_shutdown_{ false };
```
corot: This is C++11,,,  we didn't agree yet on bumping the standard, right?

corot: @spuetz, @RainerKuemmerle, unless you raise any concern, we will move master to C++17, setting the standard on the metapackage

corot (approved):

mbf_abstract_nav/include/mbf_abstract_nav/move_base_action.h
renan028: I am changing following the file convention since this is not required here

PR #321, opened 2023-07-13
Velocity Command Ignored Checking

cli18:
Problem Summary:
When the robot is disabled/not moving due to hardware/firmware issues or obstacles, and the planner keeps feeding non-zero velocity commands to the robot, we need to inform the users know something is wrong with the robot that it is ignoring cmd_vel. 

Feature:
This is an implementation of the velocity ignored checking and error displaying feature. When the robot is not moving for N seconds(set by a parameter) while the planner keeps publishing non-zero commands, it will output an error and fatal message on the ROS console and sends a error code. Velocity command ignored check is performed every time we got an new velocity command from the planner.

If user would like to enable the feature, change the `cmd_vel_ignored_tolerance` parameter to be higher than zero. Otherwise, set a value below zero.


mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
```diff
@@ -234,6 +243,9 @@ namespace mbf_abstract_nav
     //! The time the controller responded with a success output (output < 10).
     ros::Time last_valid_cmd_time_;
 
+    //The first time when the actual robot velocity is abnormal comparing to the velocity command
```
corot: When modifying existing code, follow the same comment and code stiles
```suggestion
    //! The first time when the actual robot velocity is abnormal comparing to the velocity command
```

mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
```diff
@@ -364,6 +376,9 @@ namespace mbf_abstract_nav
 
     //! replaces parameter angle_tolerance_ for the action
     double action_angle_tolerance_;
+
+    int disabled_tolerance_sec_;
```
corot: Use double for times; u don't know in advance the usage people will do
And normally you don't add units to a variable name
```suggestion
    double disabled_tolerance_sec_;
```

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -185,6 +186,41 @@ void AbstractControllerExecution::setVelocityCmd(const geometry_msgs::TwistStamp
...
+  // linear total and angular total
+  double robot_linear_total;
+  double robot_angular_total;
+
+  double cmd_linear_total;
+  double cmd_angular_total;
+
+  // compute linear velocity magnitude
+  robot_linear_total = sqrt(pow(robot_velocity.twist.linear.x, 2.0) + pow(robot_velocity.twist.linear.y, 2.0) +
+                            pow(robot_velocity.twist.linear.z, 2.0));
```
siferati:
```suggestion
  robot_linear_total = std::hypot(robot_velocity.twist.linear.x, robot_velocity.twist.linear.y);
```
F: use `std::hypot` for the other places too

mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
```diff
@@ -364,6 +376,13 @@ namespace mbf_abstract_nav
 
     //! replaces parameter angle_tolerance_ for the action
     double action_angle_tolerance_;
+
+    //! time duration for checking if the robot is ignoring cmd_vel
+    double robot_ignore_check_tolerance_;
+
+    //! robot ignore velocity feature enable/disable
+    bool robot_ignore_vel_enabled_;
```
siferati: F: Remove from class attr since not needed here

mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
```diff
@@ -219,6 +220,14 @@ namespace mbf_abstract_nav
      */
     void setVelocityCmd(const geometry_msgs::TwistStamped &vel_cmd_stamped);
 
+    /**
+     * @brief Check if the robot's actual velocity is ignoring the cmd_vel
+     * @param robot_stopped true if robot is stopped, false otherwise
+     * @param cmd_velocity the commanded velocity being published to the robot
+     * @return true if there the ignoring time exceeded tolerance, false otherwise
+     */
+    bool checkVelocityIgnore(const bool& robot_stopped, const geometry_msgs::TwistStamped& cmd_velocity);
```
siferati:
```suggestion
    bool checkVelocityIgnore(const geometry_msgs::Twist& cmd_vel);
```
S1: Check if the robot is stopped inside the function rather than passing the bool from outside
S2: Pass the twist without the stamp since you're not using it

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -72,6 +72,7 @@ AbstractControllerExecution::AbstractControllerExecution(
   private_nh.param("dist_tolerance", dist_tolerance_, 0.1);
   private_nh.param("angle_tolerance", angle_tolerance_, M_PI / 18.0);
   private_nh.param("tf_timeout", tf_timeout_, 1.0);
+  private_nh.param("ignore_check_tolerance", robot_ignore_check_tolerance_, 5.0);
```
siferati:
```suggestion
  private_nh.param("ignore_check_tolerance", ignore_check_tolerance_, 5.0);
```
S: as a general rule try to keep the same name for params and variables

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -185,6 +186,51 @@ void AbstractControllerExecution::setVelocityCmd(const geometry_msgs::TwistStamp
...
+  if (robot_stopped)
+  {
+    // linear total and angular
+    double cmd_linear;
+    double cmd_angular;
+
+    // compute linear velocity magnitude
+    cmd_linear = hypot(cmd_velocity.twist.linear.x, cmd_velocity.twist.linear.y);
+    // compute angular velocity magnitude
+    cmd_angular = fabs(cmd_velocity.twist.angular.z);
```
siferati:
```suggestion
    // compute linear velocity magnitude
    const double cmd_linear = std::hypot(cmd_velocity.twist.linear.x, cmd_velocity.twist.linear.y);
    
    // compute angular velocity magnitude
    const double cmd_angular = std::abs(cmd_velocity.twist.angular.z);
```
S: declare as const and init in same line
F: use `std::hypot` and `std::abs` instead

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -185,6 +186,51 @@ void AbstractControllerExecution::setVelocityCmd(const geometry_msgs::TwistStamp
...
+
+    if (cmd_linear > 0.01 || cmd_angular > 0.01)  // if cmd_vel is non-zero
+    {
+      // output robot ignored velocity command warning message
+      double ignored_duration = (ros::Time::now() - first_ignored_time_).toSec();
+      ROS_WARN_THROTTLE(1, "Robot ignores velocity command for %.2f seconds", ignored_duration);
```
siferati: S: Maybe add the cmd that is being ignored to the log msg

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -185,6 +186,51 @@ void AbstractControllerExecution::setVelocityCmd(const geometry_msgs::TwistStamp
...
+      first_ignored_time_ = ros::Time::now();
+    }
+  }
+  else  // robot is moving, not need to check
+  {
+    ROS_DEBUG("Robot ignore velocity command check is ok");
+    first_ignored_time_ = ros::Time::now();
+  }
+
+  return ignored_exceed_tolerance;
```
siferati: S: Avoid doing nested if statements. It's more readable to do something like this:

```python
if is_robot_moving or is_cmd_vel_zero:
    return false

# do main logic here ...
```

mbf_abstract_nav/src/controller_action.cpp
```diff
@@ -310,6 +310,13 @@ void ControllerAction::runImpl(GoalHandle &goal_handle, AbstractControllerExecut
         goal_handle.setAborted(result, result.message);
         break;
 
+      case AbstractControllerExecution::ROBOT_DISABLED:
+        ROS_FATAL_STREAM_NAMED(name_, "Robot stopped and ignored velocity commands!");
+        controller_active = false;
+        fillExePathResult(mbf_msgs::ExePathResult::STOPPED,"Robot disabled and ignored velocity commands: ",result);
```
siferati: F: `STOPPED` means the execution of the controllers was forcibly stopped (e.g. because patience ran out) -- this doesn't apply here. A better status for this case is probably `ROBOT_STUCK`

mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
```diff
@@ -219,6 +220,14 @@ namespace mbf_abstract_nav
      */
     void setVelocityCmd(const geometry_msgs::TwistStamped &vel_cmd_stamped);
 
+    /**
+     * @brief Check if the robot's actual velocity is ignoring the cmd_vel
+     * @param robot_stopped true if robot is stopped, false otherwise
+     * @param cmd_velocity the commanded velocity being published to the robot
+     * @return true if there the ignoring time exceeded tolerance, false otherwise
```
siferati:
```suggestion
     * @return true if velocity is being ignored longer than tolerance
```
F: Some of the comments' grammar throughout the code are rather questionable -- please re-read and rephrase them (feel free to also rephrase my suggestion above)

mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
```diff
@@ -219,6 +220,13 @@ namespace mbf_abstract_nav
      */
     void setVelocityCmd(const geometry_msgs::TwistStamped &vel_cmd_stamped);
 
+    /**
+     * @brief Check if the robot is ignoring the cmd_vel longer than threshold time
+     * @param cmd_velocity the latest cmd_vel being published to the robot
+     * @return true if cmd_vel is being ignored by the robot longer than tolerance time, false otherwise
+     */
+    bool checkVelocityIgnore(const geometry_msgs::Twist& cmd_velocity);
```
corot: S: a bit more precise;  and `cmd_vel` is the conventional name we use
```suggestion
    bool checkCmdVelIgnored(const geometry_msgs::Twist& cmd_vel);
```

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -72,6 +72,7 @@ AbstractControllerExecution::AbstractControllerExecution(
   private_nh.param("dist_tolerance", dist_tolerance_, 0.1);
   private_nh.param("angle_tolerance", angle_tolerance_, M_PI / 18.0);
   private_nh.param("tf_timeout", tf_timeout_, 1.0);
+  private_nh.param("robot_ignore_check_tolerance", robot_ignore_check_tolerance_, 5.0);
```
corot:
```suggestion
  private_nh.param("cmd_vel_ignored_tolerance", cmd_vel_ignored_tolerance_, 5.0);
```

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -185,6 +186,43 @@ void AbstractControllerExecution::setVelocityCmd(const geometry_msgs::TwistStamp
   // TODO so there should be no loss of information in the feedback stream
 }
 
+bool AbstractControllerExecution::checkVelocityIgnore(const geometry_msgs::Twist& cmd_velocity)
```
corot:
```suggestion
bool AbstractControllerExecution::checkVelocityIgnore(const geometry_msgs::Twist& cmd_vel)
```

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -185,6 +186,43 @@ void AbstractControllerExecution::setVelocityCmd(const geometry_msgs::TwistStamp
...
+{
+  bool ignored_exceed_tolerance = false;
```
siferati:
```suggestion
```
S: remove this variable

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -185,6 +186,43 @@ void AbstractControllerExecution::setVelocityCmd(const geometry_msgs::TwistStamp
...
+
+  bool robot_stopped = robot_info_.isRobotStopped(1e-3, 1e-3);
```
siferati:
```suggestion
  const bool robot_stopped = robot_info_.isRobotStopped(1e-3, 1e-3);
```
S: use const whenever you don't change the value after

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -185,6 +186,43 @@ void AbstractControllerExecution::setVelocityCmd(const geometry_msgs::TwistStamp
...
+
+  // compute linear and angular velocity magnitude
+  const double cmd_linear = std::hypot(cmd_velocity.linear.x, cmd_velocity.linear.y);
+  const double cmd_angular = std::abs(cmd_velocity.angular.z);
+
+  bool cmd_is_not_zero = (cmd_linear > 0.01 || cmd_angular > 0.01);
```
siferati:
```suggestion
  const bool is_cmd_zero = (cmd_linear < 0.01 && cmd_angular < 0.01);
```
S: const
F: rename, otherwise you're doing a double negative in the if statement below which is confusing (don't forget to update the if statement to match the new meaning)

```cpp
if (!robot_stopped || is_cmd_zero)
```

mbf_abstract_nav/src/controller_action.cpp
```diff
@@ -310,6 +310,13 @@ void ControllerAction::runImpl(GoalHandle &goal_handle, AbstractControllerExecut
         goal_handle.setAborted(result, result.message);
         break;
 
+      case AbstractControllerExecution::ROBOT_DISABLED:
+        ROS_FATAL_STREAM_NAMED(name_, "Robot stopped and ignored velocity commands!");
+        controller_active = false;
+        fillExePathResult(mbf_msgs::ExePathResult::ROBOT_STUCK,"Robot disabled and ignored velocity commands: ",result);
```
corot:
```suggestion
        fillExePathResult(mbf_msgs::ExePathResult::ROBOT_STUCK,"Robot stopped and ignored velocity commands: ",result);
```
Maybe better not to say "stopped",,,  can be that it never moved on the first place!!!   What about?
"Robot ignoring velocity commands for more than N seconds"  ( u need to read the param here again;  but don't do if it looks ugly)

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -376,6 +419,12 @@ void AbstractControllerExecution::run()
           vel_pub_.publish(cmd_vel_stamped.twist);
           last_valid_cmd_time_ = ros::Time::now();
           retries = 0;
+          // check if robot ignores velocity command
+          if (robot_ignore_vel_enabled && checkVelocityIgnore(cmd_vel_stamped.twist))
```
corot:
```suggestion

          // check if robot is ignoring velocity commands
          if (robot_ignore_vel_enabled && checkVelocityIgnore(cmd_vel_stamped.twist))
```

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -278,7 +316,11 @@ void AbstractControllerExecution::run()
   last_valid_cmd_time_ = ros::Time();
   int retries = 0;
   int seq = 0;
+  first_ignored_time_ = ros::Time::now();
 
+  // check if the velocity ignore check is enabled or not
+  bool robot_ignore_vel_enabled = (robot_ignore_check_tolerance_ > 0.0);
```
corot: S: don't need ()

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -185,6 +186,43 @@ void AbstractControllerExecution::setVelocityCmd(const geometry_msgs::TwistStamp
...
+                    "Robot ignores velocity command for %.2f seconds.\n The cmd_vel being ignored is: (x=%.2f, "
+                    "y=%.2f, w=%.2f)",
+                    ignored_duration, cmd_velocity.linear.x, cmd_velocity.linear.y, cmd_velocity.angular.z);
+
+  if (ignored_duration > robot_ignore_check_tolerance_)
+  {
+    // the robot is ignoring the velocity command more the threshold time
+    ROS_ERROR("Robot is disabled and the time it ignored velocity command exceeded the tolerance time: %.2f seconds",
+              robot_ignore_check_tolerance_);
+    ignored_exceed_tolerance = true;  // return true
```
siferati:
```suggestion
    return true;
```

mbf_abstract_nav/src/abstract_controller_execution.cpp
corot: S: this is forcing u to make a double negation on the if below  (avoid contorted logic whenever possible)

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -185,6 +186,43 @@ void AbstractControllerExecution::setVelocityCmd(const geometry_msgs::TwistStamp
...
+  }
+
+  return ignored_exceed_tolerance;
```
siferati:
```suggestion
  return false;
```

mbf_abstract_nav/src/abstract_controller_execution.cpp
siferati:
```suggestion
```
S1: you don't really need this variable here, just check directly inside the if statement below for `robot_ignore_check_tolerance_ > 0.0`

S2: move this logic inside `checkVelocityIgnore`. So right at the top of the function you can check if the tolerance is set to > 0 and if not just immediately return false

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -185,6 +186,43 @@ void AbstractControllerExecution::setVelocityCmd(const geometry_msgs::TwistStamp
...
+  if (!robot_stopped || !cmd_is_not_zero)
+  {
+    first_ignored_time_ = ros::Time::now();
+    return false;
+  }
+
+  // check if robot ignores the cmd_vel
+  double ignored_duration = (ros::Time::now() - first_ignored_time_).toSec();
+  ROS_WARN_THROTTLE(1,
+                    "Robot ignores velocity command for %.2f seconds.\n The cmd_vel being ignored is: (x=%.2f, "
```
corot:
```suggestion
                    "Robot ignores velocity command for %.2f seconds (commanded velocity: x=%.2f, "
```

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -185,6 +186,43 @@ void AbstractControllerExecution::setVelocityCmd(const geometry_msgs::TwistStamp
...
+                    "y=%.2f, w=%.2f)",
+                    ignored_duration, cmd_velocity.linear.x, cmd_velocity.linear.y, cmd_velocity.angular.z);
+
+  if (ignored_duration > robot_ignore_check_tolerance_)
+  {
+    // the robot is ignoring the velocity command more the threshold time
+    ROS_ERROR("Robot is disabled and the time it ignored velocity command exceeded the tolerance time: %.2f seconds",
```
corot: F: we don't know why is ignoring vel commands, so better not to make statements like "robot is disabled"  can be off, can be blocked against a wall,,,,  we don't know

mbf_abstract_nav/src/controller_action.cpp
```diff
@@ -310,6 +310,13 @@ void ControllerAction::runImpl(GoalHandle &goal_handle, AbstractControllerExecut
         goal_handle.setAborted(result, result.message);
         break;
 
+      case AbstractControllerExecution::ROBOT_DISABLED:
+        ROS_FATAL_STREAM_NAMED(name_, "Robot stopped and ignored velocity commands!");
```
siferati:
```suggestion
        ROS_FATAL_STREAM_NAMED(name_, "Robot is stopped and ignoring velocity commands!");
```
S: phrasing (same comment for the next line below)

corot (changes requested):
much better, but still needs some work

general comment:  I think here "is ignoring" makes more sense than "ignores"  (but u live in USA, so i trust your criteria)

mbf_abstract_nav/src/controller_action.cpp
cli18: Tiago said that it would be ugly if I want to read the param here. So probably it would be good to not do it here.

mbf_abstract_nav/src/controller_action.cpp
cli18: I changed it to "Robot ignored velocity commands for more than tolerance time". It makes more sense to me to write it in past tense since it already happened for N seconds.

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -185,6 +186,47 @@ void AbstractControllerExecution::setVelocityCmd(const geometry_msgs::TwistStamp
...
+    return false;
+  }
+
+  const bool robot_stopped = robot_info_.isRobotStopped(1e-3, 1e-3);
+
+  // compute linear and angular velocity magnitude
+  const double cmd_linear = std::hypot(cmd_vel.linear.x, cmd_vel.linear.y);
+  const double cmd_angular = std::abs(cmd_vel.angular.z);
+
+  const bool cmd_is_zero = (cmd_linear < 0.01 && cmd_angular < 0.01);
```
corot: S: maybe use 1e-3 here too?  
S: I think u can remove the brackets (not 100% sure)

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -185,6 +186,47 @@ void AbstractControllerExecution::setVelocityCmd(const geometry_msgs::TwistStamp
...
+                    "Robot is ignoring velocity command for %.2f seconds. (Commanded velocity: x=%.2f, "
+                    "y=%.2f, w=%.2f)",
+                    ignored_duration, cmd_vel.linear.x, cmd_vel.linear.y, cmd_vel.angular.z);
+
+  if (ignored_duration > cmd_vel_ignored_tolerance_)
+  {
+    // the robot is ignoring the velocity command more the threshold time
+    ROS_ERROR("Robot is ignoring velocity command for more than the tolerance time: %.2f seconds",
+              cmd_vel_ignored_tolerance_);
+    return true;  // return true
```
corot:
```suggestion
    return true;
```

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -185,6 +186,47 @@ void AbstractControllerExecution::setVelocityCmd(const geometry_msgs::TwistStamp
...
+  const double ignored_duration = (ros::Time::now() - first_ignored_time_).toSec();
+  ROS_WARN_THROTTLE(1,
+                    "Robot is ignoring velocity command for %.2f seconds. (Commanded velocity: x=%.2f, "
+                    "y=%.2f, w=%.2f)",
+                    ignored_duration, cmd_vel.linear.x, cmd_vel.linear.y, cmd_vel.angular.z);
+
+  if (ignored_duration > cmd_vel_ignored_tolerance_)
+  {
+    // the robot is ignoring the velocity command more the threshold time
+    ROS_ERROR("Robot is ignoring velocity command for more than the tolerance time: %.2f seconds",
```
corot, line 222: S:
```suggestion
    ROS_ERROR("Robot is ignoring velocity command for more than %.2f seconds",
```

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -278,7 +320,9 @@ void AbstractControllerExecution::run()
   last_valid_cmd_time_ = ros::Time();
   int retries = 0;
   int seq = 0;
+  first_ignored_time_ = ros::Time::now();
 
+
```
corot: extra line?

mbf_abstract_nav/src/abstract_controller_execution.cpp
cli18: I tested without the brackets and it was working fine. So I removed it.

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -185,6 +186,63 @@ void AbstractControllerExecution::setVelocityCmd(const geometry_msgs::TwistStamp
   // TODO so there should be no loss of information in the feedback stream
 }
 
+bool AbstractControllerExecution::checkCmdVelIgnored(const geometry_msgs::Twist& cmd_vel)
+{
+  // check if the velocity ignore check is enabled or not
```
corot:
```suggestion
  // check if the velocity ignored check is enabled or not
```

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -185,6 +186,63 @@ void AbstractControllerExecution::setVelocityCmd(const geometry_msgs::TwistStamp
...
+  if (cmd_vel_ignored_tolerance_ <= 0.0)
+  { 
+    // set to zero if not already zero
+    if (!first_ignored_time_.is_zero())
+    {
+      first_ignored_time_ = ros::Time();
+    }
```
corot: F: not needed if we are not checking vel ignored

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -185,6 +186,63 @@ void AbstractControllerExecution::setVelocityCmd(const geometry_msgs::TwistStamp
...
+  // compute linear and angular velocity magnitude
+  const double cmd_linear = std::hypot(cmd_vel.linear.x, cmd_vel.linear.y);
+  const double cmd_angular = std::abs(cmd_vel.angular.z);
+
+  const bool cmd_is_zero = cmd_linear < 1e-3 && cmd_angular < 1e-3;
+
+  // velocity is not being ignored
+  if (!robot_stopped || cmd_is_zero)
+  {
+    // set to zero if not already zero
```
corot: F: this comment says the same as the code.  Write comments only to clarify non-obvious steps

corot: same goes for the next 2 comments

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -185,6 +186,63 @@ void AbstractControllerExecution::setVelocityCmd(const geometry_msgs::TwistStamp
...
+  }
+
+  // check if first_ignored_time_ is zero or not
+  if (first_ignored_time_.is_zero())
+  {
+    // set first_ignored_time_ to now if it was zero
+    first_ignored_time_ = ros::Time::now();
+  }
+
+  // check if robot ignores the cmd_vel
```
corot: F: comment is wrong.  you already know that the robot is ignoring

mbf_abstract_nav/src/controller_action.cpp
```diff
@@ -310,6 +310,14 @@ void ControllerAction::runImpl(GoalHandle &goal_handle, AbstractControllerExecut
         goal_handle.setAborted(result, result.message);
         break;
 
+      case AbstractControllerExecution::ROBOT_DISABLED:
+        ROS_FATAL_STREAM_NAMED(name_, "Robot ignored velocity commands for more than tolerance time");
```
corot: F: You will have almost the same error log twice

corot (changes requested):
almost there...

mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
```diff
@@ -234,6 +242,9 @@ namespace mbf_abstract_nav
     //! The time the controller responded with a success output (output < 10).
     ros::Time last_valid_cmd_time_;
 
+    //! The first time when the robot is ignoring velocity command
```
corot:
```suggestion
    //! The time when the robot started ignoring velocity commands
```

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -185,6 +186,55 @@ void AbstractControllerExecution::setVelocityCmd(const geometry_msgs::TwistStamp
...
+
+  // compute linear and angular velocity magnitude
+  const double cmd_linear = std::hypot(cmd_vel.linear.x, cmd_vel.linear.y);
+  const double cmd_angular = std::abs(cmd_vel.angular.z);
+
+  const bool cmd_is_zero = cmd_linear < 1e-3 && cmd_angular < 1e-3;
+
+  // velocity is not being ignored
+  if (!robot_stopped || cmd_is_zero)
+  {
```
corot:
```suggestion
  if (!robot_stopped || cmd_is_zero)
  {
    // velocity is not being ignored
```

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -185,6 +186,55 @@ void AbstractControllerExecution::setVelocityCmd(const geometry_msgs::TwistStamp
...
+  }
+
+  if (first_ignored_time_.is_zero())
+  {
+    // set first_ignored_time_ to now if it was zero
+    first_ignored_time_ = ros::Time::now();
+  }
+
+  const double ignored_duration = (ros::Time::now() - first_ignored_time_).toSec();
+  ROS_WARN_THROTTLE(1,
```
corot: S: maybe this should go after the error (line 234) so you don't show both WARN and ERROR together

corot (approved):
good!

I added some extra suggestion

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -185,6 +186,56 @@ void AbstractControllerExecution::setVelocityCmd(const geometry_msgs::TwistStamp
...
+  {
+    // set first_ignored_time_ to now if it was zero
+    first_ignored_time_ = ros::Time::now();
+  }
+
+  const double ignored_duration = (ros::Time::now() - first_ignored_time_).toSec();
+
+  if (ignored_duration > cmd_vel_ignored_tolerance_)
+  {
+    // the robot is ignoring the velocity command more the threshold time
```
siferati: S: Remove this comment since it's just noise

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -185,6 +186,56 @@ void AbstractControllerExecution::setVelocityCmd(const geometry_msgs::TwistStamp
...
+      first_ignored_time_ = ros::Time();
+    }
+    return false;
+  }
+
+  if (first_ignored_time_.is_zero())
+  {
+    // set first_ignored_time_ to now if it was zero
+    first_ignored_time_ = ros::Time::now();
+  }
```
siferati, line 216:
```suggestion
```

F: Remove since we will never enter this if statement (`first_ignored_time_` is set in `AbstractControllerExecution::run` before calling this function)

mbf_abstract_nav/src/abstract_controller_execution.cpp
```diff
@@ -185,6 +186,56 @@ void AbstractControllerExecution::setVelocityCmd(const geometry_msgs::TwistStamp
...
+
+  const bool cmd_is_zero = cmd_linear < 1e-3 && cmd_angular < 1e-3;
+
+  if (!robot_stopped || cmd_is_zero)
+  {
+    // velocity is not being ignored
+    if (!first_ignored_time_.is_zero())
+    {
+      first_ignored_time_ = ros::Time();
+    }
```
siferati:
```suggestion
    first_ignored_time_ = ros::Time();
```
S: Remove the if statement since it's not doing anything -- we always enter it.

mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
```diff
@@ -234,6 +242,9 @@ namespace mbf_abstract_nav
     //! The time the controller responded with a success output (output < 10).
     ros::Time last_valid_cmd_time_;
 
+    //! The time when the robot started ignoring velocity commands
+    ros::Time first_ignored_time_;
```
siferati, line 246: F: Please rename this variable. Right now it's called `first_ignored_time_` but it's actually storing the last time a command was NOT ignored. A better name would be something like `last_followed_cmd_time_` or something similar

mbf_abstract_nav/src/abstract_controller_execution.cpp
cli18: It is not always the situation that we will enter this if statement. When cmd_vel is not being ignored last time and also this time, first_ignored_time_ is already set to zero, so it would not enter this if statement.

mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
cli18: It is actually storing the first time that the cmd_vel is being ignored by the robot. It will be set zero when the robot doesn't ignore cmd_vel. It keeps track of the first time robot ignored the cmd_vel, and it will be used to calculate the ignored_duration when the robot kept ignoring the cmd_vel until tolerance exceeded. Once the robot no longer ignores the cmd_vel, it is being set to zero again.

mbf_abstract_nav/src/abstract_controller_execution.cpp
cli18: This logic is setting the first time robot ignored the cmd_vel. It will be used to calculate ignored_duration when the robot keeps ignoring cmd_vel. So I need this here since this variable could be reset to zero if robot doesn't ignore cmd_vel in the previous loop

mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
siferati: Ah, I see. I thought you were always calling `ros::Time::now()` but you're actually calling `ros::Time()` and thus resetting to zero. Seems ok then!

mbf_abstract_nav/src/abstract_controller_execution.cpp
siferati: Yes, you're right. I didn't notice you were calling `ros::Time()` instead of `ros::Time::now()`.

However, I would still keep my suggestion here, for a different reason now.
There is no need to check if the time is zero -- you can just always reset it.
That makes the code easier to read.

siferati (approved):

PR #322, opened 2023-07-20
Change to use supercover

renan028:
# Description
As discussed here: https://github.com/magazino/move_base_flex/pull/317#discussion_r1228850967

Bresenham line drawing misses grid squares that the ray passes through. The `supercover` algorithm gets all the squares and it is the preferred algorithm to use when we search for possible collisions . 
Great description in Realtime Collision Detection "7.4.2 Uniform Grid Intersection Test".

image

In this PR, we are replacing `getLineCells` (Bresenham) by `supercover`, which is a more conservative approach.

Note that we needed to change some tests because the resolution of the map is 1.0 and the robot size is 1.0 x 1.0, thus diagonals should cover all cells around with new `supercover`.

### DIFF
--- a/mbf_costmap_nav/src/mbf_costmap_nav/footprint_helper.cpp
+++ b/mbf_costmap_nav/src/mbf_costmap_nav/footprint_helper.cpp
@@ -314,7 +314,7 @@ std::vector<Cell> FootprintHelper::getFootprintCells(double x, double y, double 
       return clearAndReturn(footprint_cells);
     }
 
-    getLineCells(x0, x1, y0, y1, footprint_cells);
+    supercover(x0, x1, y0, y1, footprint_cells);
   }
 
   //we need to close the loop, so we also have to raytrace from the last pt to first pt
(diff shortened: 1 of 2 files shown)


corot (approved):
:+1: because
- supercover result is more predictable and intuitive
- matches `find_valid_pose` behavior

RainerKuemmerle (approved):

ChristofDubs (approved):

renan028:
@Cakem1x any plans on merging this?

amock:
Hi! I took over maintenance from @Cakem1x. However, this change was introduced in the ROS 1 version of MBF, while I only maintain the ROS 2 version.

It might be best to ask the ROS 1 maintainers about this. Maybe @corot can help here?

renan028:
> Hi! I took over maintenance from @Cakem1x. However, this change was introduced in the ROS 1 version of MBF, while I only maintain the ROS 2 version.
> 
> It might be best to ask the ROS 1 maintainers about this. Maybe @corot can help here?

Sure, I will ask Jorge

PR #327, opened 2023-08-28
Change to test goal pose and check tolerance for start cell

renan028:
# Description
1) Change to test the goal pose, instead of the center of the cell which the goal pose belongs to
2) Check the tolerance for the start cell, i.e., check if the difference between start cell center and goal pose is within tolerance

Added tests for those scenarios

### DIFF
--- a/mbf_costmap_nav/include/mbf_costmap_nav/free_pose_search.h
+++ b/mbf_costmap_nav/include/mbf_costmap_nav/free_pose_search.h
@@ -207,7 +207,7 @@ private:
   bool isPoseValid(const unsigned char cost) const;
 
   /**
-   * @brief Check if state is collision-free; criteria: state free or inscribed
+   * @brief Check if state is non-lethal; criteria: state free or inscribed
    * @param state
    * @return true if state is free or inscribed
    */
--- a/mbf_costmap_nav/include/mbf_costmap_nav/free_pose_search.h
+++ b/mbf_costmap_nav/include/mbf_costmap_nav/free_pose_search.h
@@ -197,6 +197,21 @@ public:
    * @return A search solution for the given pose: best pose, state and cost
    */
   virtual SearchSolution search() const;
+
+private:
+  /**
+   * @brief Check if pose is collision-free; criteria: cost is not LETHAL, INSCRIBED and NO_INFORMATION
+   * @param cost
+   * @return true if cost is not LETHAL, INSCRIBED and NO_INFORMATION
+   */
+  bool isPoseValid(const unsigned char cost) const;
+
+  /**
+   * @brief Check if state is collision-free; criteria: state free or inscribed
+   * @param state
+   * @return true if state is free or inscribed
+   */
+  bool isStateValid(const std::uint8_t state) const;
 };
 
 } /* namespace mbf_costmap_nav */
--- a/mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
+++ b/mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
@@ -250,13 +250,9 @@ SearchSolution FreePoseSearch::search() const
   goal_cell.cost = costmap2d->getCost(goal_cell.x, goal_cell.y);
   costmap2d->mapToWorld(goal_cell.x, goal_cell.y, goal_cell_pose.x, goal_cell_pose.y);
 
-  bool test_goal_pose{ true };
   unsigned int dummy_x, dummy_y;
-  if (!costmap2d->worldToMap(config_.goal.x, config_.goal.y, dummy_x, dummy_y))
-  {
-    // if goal not in bounds, we don't start the search from the goal pose
-    test_goal_pose = false;
-  }
+  // don't start the search from the goal pose if goal is not within bounds
+  bool test_goal_pose = costmap2d->worldToMap(config_.goal.x, config_.goal.y, dummy_x, dummy_y);
 
   // add goal cell to queue if it is within linear tolerance
   if (std::hypot(goal_cell_pose.x - config_.goal.x, goal_cell_pose.y - config_.goal.y) <= config_.linear_tolerance)
--- a/mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
+++ b/mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
@@ -59,6 +59,17 @@ FreePoseSearch::FreePoseSearch(costmap_2d::Costmap2DROS& costmap, const SearchCo
   compare_strategy_ = std::move(compare_strategy.value_or(euclidean_compare));
 }
 
+bool FreePoseSearch::isPoseValid(const unsigned char cost) const
+{
+  return cost != costmap_2d::LETHAL_OBSTACLE && cost != costmap_2d::INSCRIBED_INFLATED_OBSTACLE &&
+         cost != costmap_2d::NO_INFORMATION;
+}
+
+bool FreePoseSearch::isStateValid(const std::uint8_t state) const
+{
+  return state == mbf_costmap_nav::SearchState::FREE || state == mbf_costmap_nav::SearchState::INSCRIBED;
+}
+
 std::vector<Cell> FreePoseSearch::getNeighbors(const costmap_2d::Costmap2D& costmap_2d, const Cell& cell)
 {
   std::vector<Cell> neighbors;
@@ -212,12 +223,19 @@ SearchSolution FreePoseSearch::findValidOrientation(const costmap_2d::Costmap2D&
 
 SearchSolution FreePoseSearch::search() const
 {
+  // restart markers
+  if (viz_)
+  {
+    viz_->deleteMarkers();
+  }
+
   const auto costmap2d = costmap_.getCostmap();
 
   // lock costmap so content doesn't change while adding cell costs
   boost::unique_lock<costmap_2d::Costmap2D::mutex_t> lock(*(costmap2d->getMutex()));
 
   std::unordered_set<int> in_queue_or_visited;
+  std::priority_queue<Cell, std::vector<Cell>, decltype(compare_strategy_)> queue(compare_strategy_);
 
   const std::vector<geometry_msgs::Point> footprint =
       safetyPadding(costmap_, config_.use_padded_fp, config_.safety_dist);
@@ -227,49 +245,41 @@ SearchSolution FreePoseSearch::search() const
   bool outside_or_unknown{ false };
 
   // initializing queue with the goal cell
+  // enforce bounds if goal is outside the map
+  // note that distance tolerance is wrt this new start cell
   Cell start;
-  if (costmap2d->worldToMap(config_.goal.x, config_.goal.y, start.x, start.y))
-  {
-    start.cost = costmap2d->getCost(start.x, start.y);
-  }
-  else
-  {
-    // enforce bounds if goal is outside the map
-    // note that distance tolerance is wrt this new start cell
-    int cell_x, cell_y;
-    costmap2d->worldToMapEnforceBounds(config_.goal.x, config_.goal.y, cell_x, cell_y);
-    start.x = static_cast<unsigned int>(cell_x);
-    start.y = static_cast<unsigned int>(cell_y);
-    start.cost = costmap2d->getCost(start.x, start.y);
-  }
-  in_queue_or_visited.insert(costmap2d->getIndex(start.x, start.y));
+  int cell_x, cell_y;
+  costmap2d->worldToMapEnforceBounds(config_.goal.x, config_.goal.y, cell_x, cell_y);
+  start.x = static_cast<unsigned int>(cell_x);
+  start.y = static_cast<unsigned int>(cell_y);
+  start.cost = costmap2d->getCost(start.x, start.y);
 
-  std::priority_queue<Cell, std::vector<Cell>, decltype(compare_strategy_)> queue(compare_strategy_);
-  queue.push(start);
-
-  // restart markers
-  if (viz_)
+  // check if the start cell is within tolerance and add it to the queue if it is
+  geometry_msgs::Pose2D start_world_pose;
+  costmap2d->mapToWorld(start.x, start.y, start_world_pose.x, start_world_pose.y);
+  if (std::hypot(start_world_pose.x - config_.goal.x, start_world_pose.y - config_.goal.y) <= config_.linear_tolerance)
   {
-    viz_->deleteMarkers();
+    queue.push(start);
+    in_queue_or_visited.insert(costmap2d->getIndex(start.x, start.y));
   }
 
-  while (!queue.empty())
-  {
-    Cell cell = queue.top();
-    queue.pop();
+  Cell test_cell = start;
+  // start the algorithm with the goal pose (not start cell)
+  sol.pose.x = config_.goal.x;
+  sol.pose.y = config_.goal.y;
 
-    costmap2d->mapToWorld(cell.x, cell.y, sol.pose.x, sol.pose.y);
-    ROS_DEBUG_STREAM_NAMED(LOGNAME.data(), "Checking Cell: (" << cell.x << ", " << cell.y << ") with pose: ("
+  while (true)
+  {
+    ROS_DEBUG_STREAM_NAMED(LOGNAME.data(), "Checking Cell: (" << test_cell.x << ", " << test_cell.y << ") with pose: ("
                                                               << sol.pose.x << ", " << sol.pose.y << ", "
                                                               << sol.pose.theta << ")");
 
     // Note: if the center of the robot is in costmap_2d::NO_INFORMATION, we don't accept it as a solution
-    if (cell.cost != costmap_2d::LETHAL_OBSTACLE && cell.cost != costmap_2d::INSCRIBED_INFLATED_OBSTACLE &&
-        cell.cost != costmap_2d::NO_INFORMATION)
+    if (isPoseValid(test_cell.cost))
     {
       const auto tested_sol = findValidOrientation(*costmap2d, footprint, sol.pose, config_, viz_);
       // if footprint is free or inscribed, we return the solution
-      if (tested_sol.search_state.state == SearchState::FREE || tested_sol.search_state.state == SearchState::INSCRIBED)
+      if (isStateValid(tested_sol.search_state.state))
       {
         ROS_DEBUG_STREAM_NAMED(LOGNAME.data(), "Found solution pose: " << tested_sol.pose.x << ", " << tested_sol.pose.y
                                                                        << ", " << tested_sol.pose.theta);
@@ -293,20 +303,19 @@ SearchSolution FreePoseSearch::search() const
       }
 
       ROS_DEBUG_STREAM_COND_NAMED(tested_sol.search_state.state == SearchState::LETHAL, LOGNAME.data(),
-                                  "Footprint in cell " << cell.x << ", " << cell.y << "; pose: (" << sol.pose.x << ", "
-                                                       << sol.pose.y << ", " << sol.pose.theta
+                                  "Footprint in cell " << test_cell.x << ", " << test_cell.y << "; pose: ("
+                                                       << sol.pose.x << ", " << sol.pose.y << ", " << sol.pose.theta
                                                        << ") is in lethal obstacle or unknown; skipping");
     }
-
     else
     {
-      ROS_DEBUG_STREAM_NAMED(LOGNAME.data(), "Cell: (" << cell.x << ", " << cell.y << ") with pose: (" << sol.pose.x
-                                                       << ", " << sol.pose.y << ", " << sol.pose.theta
+      ROS_DEBUG_STREAM_NAMED(LOGNAME.data(), "Cell: (" << test_cell.x << ", " << test_cell.y << ") with pose: ("
+                                                       << sol.pose.x << ", " << sol.pose.y << ", " << sol.pose.theta
                                                        << ") is in lethal obstacle or in unknown space; skipping");
     }
 
     // adding neighbors to queue
-    const std::vector<Cell> neighbors = getNeighbors(*costmap2d, cell);
+    const std::vector<Cell> neighbors = getNeighbors(*costmap2d, test_cell);
     for (const Cell& neighbor : neighbors)
     {
       int cell_index = costmap2d->getIndex(neighbor.x, neighbor.y);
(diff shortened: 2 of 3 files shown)


mbf_costmap_nav/test/free_pose_search_test.cpp
```diff
@@ -488,6 +488,75 @@ TEST_F(SearchHelperTest, enforce_bounds)
...
+  */
+
+  SearchConfig config{ M_PI_4, M_PI, 5.0, false, 0.0, toPose2D(1.345, 6.66, 0) };
+  FreePoseSearch sh(cm, config, std::nullopt, viz);
+
+  auto sol = sh.search();
+  EXPECT_EQ(sol.search_state.state, SearchState::FREE);
+  EXPECT_NEAR(sol.pose.x, 1.345, 1e-6);
+  EXPECT_NEAR(sol.pose.y, 6.66, 1e-6);
+  EXPECT_NEAR(sol.pose.theta, 0, 1e-6);
```
ChristofDubs: Do these test also pass when changing these 3 checks to `EXPECT_EQ`? If not, can you add a test with tolerance set to `0` and add such a test?

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
```diff
@@ -329,7 +346,16 @@ SearchSolution FreePoseSearch::search() const
...
-  }
+
+    if (queue.empty())
+    {
+      break;
+    }
+
+    test_cell = queue.top();
+    queue.pop();
+    costmap2d->mapToWorld(test_cell.x, test_cell.y, sol.pose.x, sol.pose.y);
```
ChristofDubs: Doesn't this line overwrite the `outside_or_unknown` solution candidate saved [here](https://github.com/magazino/move_base_flex/pull/327/files?diff=unified&w=0#diff-43fb1b9fdaf229fde2a9661293a0da41b5bfba67e7843bdb79dd19ce01fdd3cbR309)?

mbf_costmap_nav/include/mbf_costmap_nav/free_pose_search.h
```diff
@@ -197,6 +197,21 @@ class FreePoseSearch
...
+private:
+  /**
+   * @brief Check if pose is collision-free; criteria: cost is not LETHAL, INSCRIBED and NO_INFORMATION
+   * @param cost
+   * @return true if cost is not LETHAL, INSCRIBED and NO_INFORMATION
+   */
+  bool isPoseValid(const unsigned char cost) const;
+
+  /**
+   * @brief Check if state is collision-free; criteria: state free or inscribed
```
siferati: S/ F: Maybe rename `collision-free` to `non-lethal`. Coming from costmap terminology usually you'd consider inscribed cost as collision, but in this case we're not checking the center so it doesn't apply.

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
```diff
@@ -59,6 +59,17 @@ FreePoseSearch::FreePoseSearch(costmap_2d::Costmap2DROS& costmap, const SearchCo
   compare_strategy_ = std::move(compare_strategy.value_or(euclidean_compare));
 }
 
+bool FreePoseSearch::isPoseValid(const unsigned char cost) const
+{
+  return cost != costmap_2d::LETHAL_OBSTACLE && cost != costmap_2d::INSCRIBED_INFLATED_OBSTACLE &&
+         cost != costmap_2d::NO_INFORMATION;
```
siferati, line 65: Q: Why do we consider states with cost inscribed as valid but poses with cost inscribed as invalid? Shouldn't they have the same meaning?

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
```diff
@@ -212,64 +223,71 @@ SearchSolution FreePoseSearch::findValidOrientation(const costmap_2d::Costmap2D&
...
   bool outside_or_unknown{ false };
 
-  // initializing queue with the goal cell
-  Cell start;
-  if (costmap2d->worldToMap(config_.goal.x, config_.goal.y, start.x, start.y))
+  // enforce bounds if goal is outside the map
+  int test_cell_x, test_cell_y;
+  costmap2d->worldToMapEnforceBounds(config_.goal.x, config_.goal.y, test_cell_x, test_cell_y);
+  test_cell.x = static_cast<unsigned int>(test_cell_x);
+  test_cell.y = static_cast<unsigned int>(test_cell_y);
```
siferati:
```suggestion
  costmap2d->worldToMapEnforceBounds(config_.goal.x, config_.goal.y, test_cell.x, test_cell.y);
```
S: why create additional vars for this? calling `worldToMapEnforceBounds` should work even with uint no?

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
```diff
@@ -212,64 +223,71 @@ SearchSolution FreePoseSearch::findValidOrientation(const costmap_2d::Costmap2D&
...
   {
-    // enforce bounds if goal is outside the map
-    // note that distance tolerance is wrt this new start cell
-    int cell_x, cell_y;
-    costmap2d->worldToMapEnforceBounds(config_.goal.x, config_.goal.y, cell_x, cell_y);
-    start.x = static_cast<unsigned int>(cell_x);
-    start.y = static_cast<unsigned int>(cell_y);
-    start.cost = costmap2d->getCost(start.x, start.y);
+    sol.pose.x = test_pose.x;
+    sol.pose.y = test_pose.y;
```
siferati: S: can't we add this to the visited set?

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
```diff
@@ -212,64 +223,71 @@ SearchSolution FreePoseSearch::findValidOrientation(const costmap_2d::Costmap2D&
...
+  test_cell.cost = costmap2d->getCost(test_cell.x, test_cell.y);
+
+  // the cell center might be different from the goal pose (depends on costmap resolution)
+  geometry_msgs::Pose2D test_pose;
+  costmap2d->mapToWorld(test_cell.x, test_cell.y, test_pose.x, test_pose.y);
+
+  int cell_no_bounds_x, cell_no_bounds_y;
+  costmap2d->worldToMapNoBounds(config_.goal.x, config_.goal.y, cell_no_bounds_x, cell_no_bounds_y);
+  // if goal in bounds, we start the search from the goal pose and continue from the goal cell's center
+  if (cell_no_bounds_x == test_cell_x && cell_no_bounds_y == test_cell_y)
```
siferati:
```suggestion
  // if goal in bounds, we start the search from the goal pose and continue from the goal cell's center
  unsigned int dummy_x, dummy_y;
  if (!costmap2d->worldToMap(config_.goal.x, config_.goal.y, dummy_x, dummy_y))
```
S: Use the version that returns a bool instead of comparing the cell coords
(why does this version receive uint but enforce bounds receives int :joy:)

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
siferati: R: I find this while loop to be very un-intuitive. You're using variables that were set from outside as the variables to evaluate. Can you refactor this to follow the standard flow? Something like:

```cpp
while (!queue.empty())
{
    Cell test_cell = queue.top();
    queue.pop();
    SearchSolution sol;
    costmap2d->mapToWorld(test_cell.x, test_cell.y, sol.pose.x, sol.pose.y);

    // ...
}
```

Remove `test_cell` and `sol` from outside the loop, and simply add the poses to the queue.
I guess you need to test the goal pose outside the loop, so create a new Cell object for that rather than re-using `test_cell`

mbf_costmap_nav/test/free_pose_search_test.cpp
renan028: yes, they do. I am changing to zero

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
renan028: you are correct. Fixed and added test for that

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
renan028: When you check the pose, you query costmap the cost as if the center of the robot is on that cell. That means you are checking the pose of the robot. If the cost is `INSCRIBED`, the robot is for sure in collision and we don't need to verify the footprint (it will be LETHAL for sure).
To check the state, you query all cells that belongs to the footprint (border and inside) and then get the max cost. Therefore, if the border is `INSCRIBED`, it is fine. The problem arises only if a cell is LETHAL.

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
renan028: a reference of type `int &` (not const-qualified) cannot be initialized with a value of type `unsigned int`

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
renan028: That was the previous loop, but I had to change to test the goal pose first. 
I didn't understand your suggestion. Are you suggesting to copy the whole while loop outside for just the initial goal pose?
I think this is bad. Another approach would be to add a flag for the first pose, which is also ugly.

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
```diff
@@ -212,64 +223,84 @@ SearchSolution FreePoseSearch::findValidOrientation(const costmap_2d::Costmap2D&
...
-  queue.push(start);
-
-  // restart markers
-  if (viz_)
-  {
-    viz_->deleteMarkers();
+    // if out of bounds, we start the search from the goal cell
+    goal_pose_tested = true;
+    queue.push(goal_cell);
+    in_queue_or_visited.insert(costmap2d->getIndex(goal_cell.x, goal_cell.y));
```
siferati: Q / S: shouldn't check if within distance like the lines above?

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
```diff
@@ -212,64 +223,84 @@ SearchSolution FreePoseSearch::findValidOrientation(const costmap_2d::Costmap2D&
...
+  geometry_msgs::Pose2D goal_cell_pose;  // this is the pose of the center of the goal cell (not the goal pose)
+  int goal_cell_x, goal_cell_y;
+  costmap2d->worldToMapEnforceBounds(config_.goal.x, config_.goal.y, goal_cell_x, goal_cell_y);
+  goal_cell.x = static_cast<unsigned int>(goal_cell_x);
+  goal_cell.y = static_cast<unsigned int>(goal_cell_y);
+  goal_cell.cost = costmap2d->getCost(goal_cell.x, goal_cell.y);
+  costmap2d->mapToWorld(goal_cell.x, goal_cell.y, goal_cell_pose.x, goal_cell_pose.y);
+
+  // if goal in bounds, we start the search from the goal pose (not goal_cell_pose) and continue from the goal cell
+  bool goal_pose_tested{ false };
```
siferati: S: Rename to `test_goal_pose` and invert the meaning

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
```diff
@@ -212,64 +223,84 @@ SearchSolution FreePoseSearch::findValidOrientation(const costmap_2d::Costmap2D&
...
+    // if out of bounds, we start the search from the goal cell
+    goal_pose_tested = true;
+    queue.push(goal_cell);
+    in_queue_or_visited.insert(costmap2d->getIndex(goal_cell.x, goal_cell.y));
   }
 
-  while (!queue.empty())
+  SearchSolution sol;
+  sol.pose.theta = config_.goal.theta;
+  SearchSolution unknow_outside_sol;
```
siferati: S: Rename to `no_info_sol`, or `unknown_sol` or something else. Having 2 words is confusing

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
renan028: change to check tolerance for pose outside bounds and add test

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
renan028: change to optional, removing the boolean

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
```diff
@@ -212,64 +223,78 @@ SearchSolution FreePoseSearch::findValidOrientation(const costmap_2d::Costmap2D&
...
+  costmap2d->mapToWorld(goal_cell.x, goal_cell.y, goal_cell_pose.x, goal_cell_pose.y);
+
+  bool test_goal_pose{ true };
+  unsigned int dummy_x, dummy_y;
+  if (!costmap2d->worldToMap(config_.goal.x, config_.goal.y, dummy_x, dummy_y))
   {
-    start.cost = costmap2d->getCost(start.x, start.y);
+    // if goal not in bounds, we don't start the search from the goal pose
+    test_goal_pose = false;
   }
```
ChristofDubs: Suggestion: Maybe just do
``` 
// don't start the search from the goal pose if goal is not within bounds
bool test_goal_pose = costmap2d->worldToMap(config_.goal.x, config_.goal.y, dummy_x, dummy_y);
```
directly

ChristofDubs: Another suggestion would be to just change the queue to poses? You have to convert all the cells to poses anyway for the tolerance check, so you might as well make a single queue of poses, without special `test_goal_pose` flag?

renan028: I don't like the second suggestion because the `queue` element shows intent and the search model. 
In our case, we are searching the free `cell` in the costmap, we update the queue according to neighboring cells, and the priority method is also an euclidean distance between cells, that could be changed to `cost` (gradient) or other custom comparison in the future. 
Therefore, changing the element is not a small change in the code (we need to change comparison function, where cost will not be available, and other parts of the code), but it also affects readability, intent and the search model. 
Besides, I don't think it is justified by the unique case of testing the goal first.

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
```diff
@@ -212,64 +223,74 @@ SearchSolution FreePoseSearch::findValidOrientation(const costmap_2d::Costmap2D&
...
-  else
+  // enforce bounds if goal is outside the map; and get the pose of the center of the cell
+  Cell goal_cell;
+  geometry_msgs::Pose2D goal_cell_pose;  // this is the pose of the center of the goal cell (not the goal pose)
+  int goal_cell_x, goal_cell_y;
+  costmap2d->worldToMapEnforceBounds(config_.goal.x, config_.goal.y, goal_cell_x, goal_cell_y);
+  goal_cell.x = static_cast<unsigned int>(goal_cell_x);
+  goal_cell.y = static_cast<unsigned int>(goal_cell_y);
+  goal_cell.cost = costmap2d->getCost(goal_cell.x, goal_cell.y);
+  costmap2d->mapToWorld(goal_cell.x, goal_cell.y, goal_cell_pose.x, goal_cell_pose.y);
```
siferati:
```suggestion
  // enforce bounds if goal is outside the map; and get the pose of the center of the cell
  int goal_cell_x, goal_cell_y;
  costmap2d->worldToMapEnforceBounds(config_.goal.x, config_.goal.y, goal_cell_x, goal_cell_y);
  Cell goal_cell = {goal_cell_x, goal_cell_y, costmap2d->getCost(goal_cell_x, goal_cell_y)};
  
  // this is the pose of the center of the goal cell (not the goal pose)
  geometry_msgs::Pose2D goal_cell_pose;
  costmap2d->mapToWorld(goal_cell.x, goal_cell.y, goal_cell_pose.x, goal_cell_pose.y);
```
S1: better spacing
S2: declare and init `goal_cell` in the same line
S3: maybe rename `goal_cell_x` to `dummy_x` -- it's confusing having both `goal_cell_x` and `goal_cell.x`. But down below you already have another dummy that's unsigned int... so maybe rename to `dummy_int_x` and `dummy_uint_x`? :laughing:

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
```diff
@@ -212,64 +223,74 @@ SearchSolution FreePoseSearch::findValidOrientation(const costmap_2d::Costmap2D&
...
+
+  unsigned int dummy_x, dummy_y;
+  // don't start the search from the goal pose if goal is not within bounds
+  bool test_goal_pose = costmap2d->worldToMap(config_.goal.x, config_.goal.y, dummy_x, dummy_y);
```
siferati:
```suggestion
  // don't start the search from the goal pose if goal is not within bounds
  unsigned int dummy_x, dummy_y;
  bool test_goal_pose = costmap2d->worldToMap(config_.goal.x, config_.goal.y, dummy_x, dummy_y);
```
S1: move comment up for better grouping and readability
S2: move these 2 lines below (after the if statement) in order for the logic around `goal_cell` to be grouped

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
```diff
@@ -212,64 +223,74 @@ SearchSolution FreePoseSearch::findValidOrientation(const costmap_2d::Costmap2D&
...
+      sol.pose.x = config_.goal.x;
+      sol.pose.y = config_.goal.y;
+      test_goal_pose = false;
+    }
+    else
+    {
+      test_cell = queue.top();
+      queue.pop();
+      costmap2d->mapToWorld(test_cell.x, test_cell.y, sol.pose.x, sol.pose.y);
+    }
```
siferati, line 282:
```suggestion
    Cell test_cell = queue.top();
    if (test_goal_pose)
    {
      sol.pose.x = config_.goal.x;
      sol.pose.y = config_.goal.y;
      test_goal_pose = false;
    }
    else
    {
      queue.pop();
      costmap2d->mapToWorld(test_cell.x, test_cell.y, sol.pose.x, sol.pose.y);
    }
```
S: if you're testing the goal pose, then the first (and only) item in the queue is the goal cell.
So we can simplify the code by reading queue.top() outside the if statement

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
```diff
@@ -283,30 +304,28 @@ SearchSolution FreePoseSearch::search() const
...
-                                                       << ", " << sol.pose.y << ", " << sol.pose.theta
+      ROS_DEBUG_STREAM_NAMED(LOGNAME.data(), "Cell: (" << test_cell.x << ", " << test_cell.y << ") with pose: ("
+                                                       << sol.pose.x << ", " << sol.pose.y << ", " << sol.pose.theta
                                                        << ") is in lethal obstacle or in unknown space; skipping");
     }
 
     // adding neighbors to queue
-    const std::vector<Cell> neighbors = getNeighbors(*costmap2d, cell);
+    const std::vector<Cell> neighbors = getNeighbors(*costmap2d, test_cell);
     for (const Cell& neighbor : neighbors)
```
siferati:
```suggestion
    for (const auto& neighbor : neighbors)
```
S1: `auto`

---

S2: replace the following if statement

https://github.com/magazino/move_base_flex/blob/c44b78e17cf0a4d4063907a6f50beea31fb61e76/mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp#L332

with

```cpp
if (in_queue_or_visited.count(cell_index))
```

This line is outside the PR so I can't comment on it :sweat_smile:

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
```diff
@@ -212,64 +223,74 @@ SearchSolution FreePoseSearch::findValidOrientation(const costmap_2d::Costmap2D&
...
-
-  // restart markers
-  if (viz_)
-  {
-    viz_->deleteMarkers();
-  }
-
-  while (!queue.empty())
+  SearchSolution sol;
+  sol.pose.theta = config_.goal.theta;
```
siferati: S: do this inside the while loop. Right now it works under the assumption that the value will never be changed inside the loop, which could change in the future

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
renan028: we don't need to use `goal_cell.x`. I am removing it and using `goal_cell_x` instead

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
renan028: In this case, they have equal performance, and the result is the same since the container is `std::unordered_set<int>`
But I prefer to use the algorithm with better intent (better expresses what I am doing).
If it was c++20, I would use `contains`
fyi, same discussion: https://stackoverflow.com/questions/14159682/unordered-map-which-one-is-faster-find-or-count

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
renan028: This is wrong. The queue can be empty and the suggestion breaks the code

mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp
siferati: I'd argue count also expresses clear intent: checking if an element exists in the container is the same as checking if there is at least 1 elem of it (i.e. count > 0). So since count is shorter / easier to read usually I prefer it over find. But it's just a suggestion so feel free to ignore :shrug:

siferati (approved):

ChristofDubs (approved):

corot (approved):

PR #329, opened 2023-09-13
Fix service zero tolerance numerical issue

renan028:
# Description

In the `callServiceFindValidPose` service we use `tf` utility functions to convert `quarternion` to `yaw` back and forth: `tf::getYaw` and `tf::createQuaternionMsgFromYaw` because of the `FreePoseSearch` class interfaces.

This can lead to numerical issues. For example, when the starting angle for the search is the solution, the original angle (service request) can be slightly different from the output (service response) because of the conversions.

Adding a check to fix the numerical issue: if the difference from input to output is less than the `ANGLE_INCREMENT`, we use the input angle without conversions.

### DIFF
--- a/mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
+++ b/mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
@@ -48,6 +48,7 @@
 #include <nav_core_wrapper/wrapper_local_planner.h>
 #include <nav_core_wrapper/wrapper_recovery_behavior.h>
 #include <xmlrpcpp/XmlRpc.h>
+#include <angles/angles.h>
 
 #include "mbf_costmap_nav/footprint_helper.h"
 #include "mbf_costmap_nav/costmap_navigation_server.h"
@@ -885,7 +886,12 @@ bool CostmapNavigationServer::callServiceFindValidPose(mbf_msgs::FindValidPose::
   response.pose.pose.position.x = sol.pose.x;
   response.pose.pose.position.y = sol.pose.y;
   response.pose.pose.position.z = 0;
-  response.pose.pose.orientation = tf::createQuaternionMsgFromYaw(sol.pose.theta);
+
+  // if the difference between solution angle and requested angle is less than ANGLE_INCREMENT,
+  // use the requested one to avoid numerical issues
+  response.pose.pose.orientation = angles::shortest_angular_distance(goal.theta, sol.pose.theta) < ANGLE_INCREMENT ?
+                                       request.pose.pose.orientation :
+                                       tf::createQuaternionMsgFromYaw(sol.pose.theta);
   response.pose.header.frame_id = costmap_frame;
   response.pose.header.stamp = ros::Time::now();
   response.state = sol.search_state.state;
--- a/mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
+++ b/mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
@@ -888,10 +888,24 @@ bool CostmapNavigationServer::callServiceFindValidPose(mbf_msgs::FindValidPose::
   response.pose.pose.position.z = 0;
 
   // if the difference between solution angle and requested angle is less than ANGLE_INCREMENT,
-  // use the requested one to avoid numerical issues
-  response.pose.pose.orientation = angles::shortest_angular_distance(goal.theta, sol.pose.theta) < ANGLE_INCREMENT ?
-                                       request.pose.pose.orientation :
-                                       tf::createQuaternionMsgFromYaw(sol.pose.theta);
+  // use the requested one to avoid violating a very small angle_tolerance (e.g. 0)
+  response.pose.pose.orientation =
+      std::abs(angles::shortest_angular_distance(goal.theta, sol.pose.theta)) < ANGLE_INCREMENT ?
+          request.pose.pose.orientation :
+          tf::createQuaternionMsgFromYaw(sol.pose.theta);
+
+  const double linear_dist = std::hypot(goal.x - sol.pose.x, goal.y - sol.pose.y);
+  const double angular_dist =
+      std::abs(angles::shortest_angular_distance(goal.theta, tf::getYaw(response.pose.pose.orientation)));
+  ROS_DEBUG_STREAM("Solution distance: " << linear_dist << ", angle: " << angular_dist);
+
+  // checking if the solution does not violate the requested distance and angle tolerance
+  if (linear_dist > request.dist_tolerance || angular_dist > request.angle_tolerance)
+  {
+    ROS_ERROR_STREAM("Solution violates requested distance and/or angle tolerance");
+    return false;
+  }
+
   response.pose.header.frame_id = costmap_frame;
   response.pose.header.stamp = ros::Time::now();
   response.state = sol.search_state.state;
--- a/mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
+++ b/mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
@@ -887,12 +887,10 @@ bool CostmapNavigationServer::callServiceFindValidPose(mbf_msgs::FindValidPose::
   response.pose.pose.position.y = sol.pose.y;
   response.pose.pose.position.z = 0;
 
-  // if the difference between solution angle and requested angle is less than ANGLE_INCREMENT,
-  // use the requested one to avoid violating a very small angle_tolerance (e.g. 0)
+  // if solution angle and requested angle are the same (after conversion),
+  // use the requested (quaternion) one to avoid violating a very small angle_tolerance (e.g. 0)
   response.pose.pose.orientation =
-      std::abs(angles::shortest_angular_distance(goal.theta, sol.pose.theta)) < ANGLE_INCREMENT ?
-          request.pose.pose.orientation :
-          tf::createQuaternionMsgFromYaw(sol.pose.theta);
+      goal.theta == sol.pose.theta ? request.pose.pose.orientation : tf::createQuaternionMsgFromYaw(sol.pose.theta);
 
   const double linear_dist = std::hypot(goal.x - sol.pose.x, goal.y - sol.pose.y);
   const double angular_dist =
--- a/mbf_costmap_nav/CMakeLists.txt
+++ b/mbf_costmap_nav/CMakeLists.txt
@@ -5,6 +5,7 @@ set(CMAKE_CXX_STANDARD 17)
 
 find_package(catkin REQUIRED
   COMPONENTS
+  angles
   costmap_2d
   dynamic_reconfigure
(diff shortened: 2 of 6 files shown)


mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
```diff
@@ -885,7 +886,12 @@ bool CostmapNavigationServer::callServiceFindValidPose(mbf_msgs::FindValidPose::
   response.pose.pose.position.x = sol.pose.x;
   response.pose.pose.position.y = sol.pose.y;
   response.pose.pose.position.z = 0;
-  response.pose.pose.orientation = tf::createQuaternionMsgFromYaw(sol.pose.theta);
+
+  // if the difference between solution angle and requested angle is less than ANGLE_INCREMENT,
+  // use the requested one to avoid numerical issues
```
ChristofDubs: `to avoid numerical issues` is a bit vague.

I'd write that it is `to avoid violating a very small angle_tolerance (e.g. 0)`

mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
```diff
@@ -885,7 +886,12 @@ bool CostmapNavigationServer::callServiceFindValidPose(mbf_msgs::FindValidPose::
...
+  response.pose.pose.orientation = angles::shortest_angular_distance(goal.theta, sol.pose.theta) < ANGLE_INCREMENT ?
+                                       request.pose.pose.orientation :
+                                       tf::createQuaternionMsgFromYaw(sol.pose.theta);
```
ChristofDubs: Suggestion: to avoid similar issues in the future, it might make sense to explicitly check here whether the angle tolerance or dist_tolerance are violated (assert and / or return false)? I.e. `std::hypot(dx, dy) <= dist_tolerance` and `abs(getYaw(response) - getYaw(request)) <= angle_tolerance`.

Another numerical issue that might happen is when the `angle_tolerance` is a multiple of `ANGLE_INCREMENT` (but that should be a separate PR). Position related issues seem unlikely, because the search already uses a hypot check before adding to the queue.

mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
```diff
@@ -885,7 +886,12 @@ bool CostmapNavigationServer::callServiceFindValidPose(mbf_msgs::FindValidPose::
   response.pose.pose.position.x = sol.pose.x;
   response.pose.pose.position.y = sol.pose.y;
   response.pose.pose.position.z = 0;
-  response.pose.pose.orientation = tf::createQuaternionMsgFromYaw(sol.pose.theta);
+
+  // if the difference between solution angle and requested angle is less than ANGLE_INCREMENT,
+  // use the requested one to avoid numerical issues
+  response.pose.pose.orientation = angles::shortest_angular_distance(goal.theta, sol.pose.theta) < ANGLE_INCREMENT ?
```
corot: F/Q: I think this can be negative

corot (changes requested):
.

mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
renan028: oh true, fixed

mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
renan028: I have added the check.

hum, I think we can have numerical issues in different places in this case.
For instance, in the search loop that may return earlier since it may not check all increments correctly (missing the last one because it is a multiple and a floating point error could make it slightly below); or the problem could be when we sum up the increments (floating point error in the sum); or in the conversion again itself. 
I think it will be hard to track all possibilities, so I would advise to use a tolerance slightly above/below increment multiples.

ChristofDubs: So I think adding the check makes sense. But also, we should avoid that this check is even necessary here in the code, because the search loop might return, and the solution gets rejected, so the service returns that there was no solution, even though there exists one that the search would have found if it didn't already terminate.

Also, what would `advising the user to use a tolerance slightly above/below increment multiples` mean? The user would have to write code like `if (std::abs(std::remainer(tolerance, increment)) < 1e-6) {increment -= 1e-6;}`?
Rather than telling the user to do this, I think we should just slightly modify the increment in the code here:

https://github.com/magazino/move_base_flex/blob/c2e7619841ab80557461443b857d1c8514e22e97/mbf_costmap_nav/src/mbf_costmap_nav/free_pose_search.cpp#L154-L157

? 

Something like:
```
const double reduced_tol = std::min(config.angle_tolerance - 1e-6, M_PI);
const int num_steps = 1 + std::max(0, std::ceil(reduced_tol / config.angle_increment));
const double increment = reduced_tol / std::max(1, num_steps - 1);
for (int i = 0; i < num_steps; ++num_steps) {
    const std::vector<double> thetas = i == 0 ? std::vector<double>{ pose_2d.theta } :
                                                   std::vector<double>{ pose_2d.theta + i*increment, pose_2d.theta - i*increment };
```

renan028: Hum, I don't think we can solve floating point issues that way.

What if the user pass: `config.angle_tolerance = config.angle_increment + 1e-6` ?

Then some conversion, or ROS comm serialization, or floating arithmetic, etc could slightly change that number anyway. Then:  

`reduced_tol ~= config.angle_increment;` (slightly below or slightly above) and then
`num_steps = 2` or `num_steps = 3` ?

**Edit**
Actually, Is this the expected behavior after the change? Maybe I understood wrong.

`config.angle_tolerance = config.angle_increment + 1e-8` (because user wants the increment check)
`reduced_tol = config.angle_increment - ε` 
`num_steps = 2`
`increment = config.angle_increment - ε`
We will test: `i = {0,1}` -> `{pose_2d.theta, pose_2d.theta + config.angle_increment - ε, pose_2d.theta + config.angle_increment + ε}` 
So we are changing to not test the increment instead?

corot (approved):

corot (commented):
.

mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
ChristofDubs: Maybe I should have clarified: In my suggestion above, the idea is to interpret the `angle_increment` as a "maximum step size", and choose an actual step size is below that, such that the `angle_tolerance` is fully covered, without violating it.

renan028: Got it. However, if I understood it correctly, we will not fully cover the `angle_tolerance`, but remove `1e-6` from it instead: https://godbolt.org/z/1jhbfY619
Therefore, if the user want to fully cover the tolerance, it will need to send `angle_tolerance + 1e-6` (I personally don't like this hardcoded value in the API, it is a bit unexpected)
And the `angle_increment` is not respected anymore, which may be unexpected for the user experience (as can be seen in the example `tol = inc + 2 * 1e-6;` it uses half of the increment)

ChristofDubs: > Got it. However, if I understood it correctly, we will not fully cover the angle_tolerance, but remove 1e-6 from it instead: https://godbolt.org/z/1jhbfY619

yes. That is done on purpose to not violate the tolerance.

I'd also say it covers the tolerance much much better than the current implementation for some scenarios? E.g.

- `tol: 1.5, inc: 0.6`:
    - now: { 0, 0.6, -0.6, 1.2, -1,2}
    - my suggestion: { 0, 0.4999996667, -0.4999996667, 0.9999993333, -0.9999993333, 1.499999, -1.499999} -> this covers the tolerance better
    
- `tol: 1.2, inc: 0.12` (https://godbolt.org/z/hYvxc63z3)
    - now: {0, ..., 1.08, -1.08}
    - my suggestion: { 0, ..., 1.199999, -1.199999} -> this covers the tolerance better

> Therefore, if the user want to fully cover the tolerance, it will need to send angle_tolerance + 1e-6  (I personally don't like this hardcoded value in the API, it is a bit unexpected)

Since we're dealing with numbers only up to `pi`, I think the `1e-6` value could be reduced to about `std::numeric_limits<double>::epsilon * M_PI`.

Either way, the small reduction of the tolerance has practically basically no effect, unlike the current version has in the two examples above (-> [good enough](https://flylib.com/books/en/1.315.1.17/1/))

> And the angle_increment is not respected anymore, which may be unexpected for the user experience (as can be seen in the example tol = inc + 2 * 1e-6; it uses half of the increment)

Yes so as written above, angle_increment is interpreted as a "maximum step size". So to make this clear to the user, I would of course recommend to rename `angle_increment` to `angle_max_step_size` or sth similar.

renan028: > I'd also say it coves the tolerance much much better than the current implementation for some scenarios

Yes, indeed. My initial idea was to left to the user to plan a step size and tolerance by themselves, but it is better to do it in the API as you suggested. By changing the name of the variable, we suggest that the algorithm will figure out a better way to handle tolerances and increments. thanks

mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
```diff
@@ -885,7 +886,26 @@ bool CostmapNavigationServer::callServiceFindValidPose(mbf_msgs::FindValidPose::
   response.pose.pose.position.x = sol.pose.x;
   response.pose.pose.position.y = sol.pose.y;
   response.pose.pose.position.z = 0;
-  response.pose.pose.orientation = tf::createQuaternionMsgFromYaw(sol.pose.theta);
+
+  // if the difference between solution angle and requested angle is less than ANGLE_INCREMENT,
+  // use the requested one to avoid violating a very small angle_tolerance (e.g. 0)
+  response.pose.pose.orientation =
+      std::abs(angles::shortest_angular_distance(goal.theta, sol.pose.theta)) < ANGLE_INCREMENT ?
```
ChristofDubs: So since `ANGLE_INCREMENT` is now max step size, this check isn't valid anymore since the search will make smaller steps internally.

Not sure how to best accommodate this. Maybe we could add a min_step_size as well, and use that here? Or return more information from `free_pose_search.search();` (e.g. by how much the goal was displaced)?

renan028: yes, the check is invalid now. 
The information of how much the goal was displaced is available to the user since he has the goal and he has the result. I think we can just check if the input == output (after conversion) and, in this case, use the requested angle (without conversion), avoiding floating erros.

ChristofDubs (approved):

corot (approved):

PR #330, opened 2023-10-31
Fix angle oscillation

renan028:
# Description
If the robot is too slow to execute a spin-in-place, oscillation will be triggered.
However, that is not an actual oscillation, since the robot is spinning to goal.

# Testing
Behavior before (`oscillation_angle == 𝝿`):

https://github.com/magazino/move_base_flex/assets/6097267/51b326b5-1b17-41ee-aeaf-0fc8e555dd83

Behavior after (`oscillation_angle == 0.1`):

https://github.com/magazino/move_base_flex/assets/6097267/13f40a9c-cd51-44f5-a29e-a0c52a8d23b7


corot (approved):
I'm realizing that the param name is not clear enough,,,  should be `oscillation_reset_distance`

siferati (approved):

PR #332, opened 2023-11-15
accept tf_timeout < 0 to ignore timeout

knorth55:
this PR change to accept `tf_timeout` param `-1` to ignore `transformation` timeout check.
this is useful when we use `tf_static_transformer` for debugging.

### DIFF
--- a/mbf_utility/src/navigation_utility.cpp
+++ b/mbf_utility/src/navigation_utility.cpp
@@ -67,7 +67,7 @@ bool getRobotPose(const TF &tf,
                                timeout,
                                local_pose,
                                robot_pose);
-  if (success && ros::Time::now() - robot_pose.header.stamp > timeout)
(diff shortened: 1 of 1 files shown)


corot:
Why not just passing a very big value?
And can u actually have a negative ros::Duration?

I would instead reconsider your use case; I'm not at all convince about the need of this change

knorth55:
`tf_static_publisher` uses the first timestamp when node starts.
we run robots for weeks, so the duration will be super large.
so i use `-1` as `tf.lookUTransform`

knorth55:
we dont use map and amcl, so we directly connect `odom` and `map` by `tf_static_publisher`.

corot:
> `tf_static_publisher` uses the first timestamp when node starts. we run robots for weeks, so the duration will be super large. so i use `-1` as `tf.lookUTransform`

Yes, static tfs only need to be published once, but tf library considers them as always available, at any timestamp.
So that should not require this PR. You may have some other hidden problem in the code

PR #333, opened 2023-11-15
add rosparam to accept node name change

knorth55:
this PR accept to change the node name for `legacy relay`.
this PR is useful when you want to keep the node name as `move_base`, because some `rviz` assumes the `move_base` node name.
I use as below and I don't need to change rviz config which I used for `ros-planning/navigation`.
- `move_base_flex_node_name`: `move_base`
- `move_base_node_name`: `move_base`

### DIFF
--- a/mbf_costmap_nav/scripts/move_base_legacy_relay.py
+++ b/mbf_costmap_nav/scripts/move_base_legacy_relay.py
@@ -115,23 +115,25 @@ def mb_reconf_cb(config, level):
 
 
 if __name__ == '__main__':
-    rospy.init_node("move_base")
+    rospy.init_node("move_base_legacy_relay")
+    node_name = rospy.get_param('~move_base_flex_node_name', 'move_base_flex')
+    move_base_node_name = rospy.get_param('~move_base_node_name', 'move_base')
 
     # TODO what happens with malformed target goal???  FAILURE  or INVALID_POSE
     # txt must be:  "Aborting on goal because it was sent with an invalid quaternion"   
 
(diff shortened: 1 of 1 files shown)


corot:
Thanks @knorth55 for the contribution, but these kind of things are much better handled with remappings
Using params to configure topic names is considered an anti-pattern and so strongly discouraged
Moreover, changing RViz config is way easier, as it's all handled internally by RViz, and so much harder to make mistakes

knorth55:
i strongly think it is not a good idea to configure rviz again, because we want to have two different rviz configure for move_base and move_base_flex.
at least, i dont want to change plenty of rviz configs we already use.

remapping for simple action server is troublesome so i use rosparam, but it is an anti pattern. you are right.

corot:
> remapping for simple action server is troublesome so i use rosparam, but it is an anti pattern. you are right.

not 100% sure, but I think it's possible to remap actions :thinking:

knorth55:
yes im sure i can.
but remapping `goal, feedback, status, cancel result` is kind of troublesome to me.
anyway this PR is trivial, so it is ok not to be merged.

corot:
> yes im sure i can. but remapping `goal, feedback, status, cancel result` is kind of troublesome to me. anyway this PR is trivial, so it is ok not to be merged.

no no,,,  I mean, remap the action name, not the individual topics  (ie move_base  -->  navigate)

knorth55:
Oh really, can I do that?
In that case, I don't need this PR.

PR #334, opened 2023-11-15
initialize last_oscillation_pose_

knorth55:
this PR fix the issue that `last_oscillation_pose_` is not initialized.
this PR change to initialize the robot pose when `started`

PR #335, opened 2023-11-15
support oscillation angle in move_base

knorth55:
this PR needs #334 

this PR adds oscillation angle feature in `move_base`.
now only `exe_path` has the `oscillation_angle` feature, but it is also useful for `move_base`, too.

### DIFF
--- a/mbf_abstract_nav/src/mbf_abstract_nav/__init__.py
+++ b/mbf_abstract_nav/src/mbf_abstract_nav/__init__.py
@@ -38,3 +38,5 @@ def add_mbf_abstract_nav_params(gen):
             "How long in seconds to allow for oscillation before executing recovery behaviors", 0.0, 0, 60)
     gen.add("oscillation_distance", double_t, 0,
             "How far in meters the robot must move to be considered not to be oscillating", 0.5, 0, 10)
+    gen.add("oscillation_angle", double_t, 0,
+            "How far in radian the robot must rotate to be considered not to be oscillating", 3.14, 0, 6.28)
--- a/mbf_abstract_nav/src/mbf_abstract_nav/__init__.py
+++ b/mbf_abstract_nav/src/mbf_abstract_nav/__init__.py
@@ -39,4 +39,4 @@ def add_mbf_abstract_nav_params(gen):
     gen.add("oscillation_distance", double_t, 0,
             "How far in meters the robot must move to be considered not to be oscillating", 0.5, 0, 10)
     gen.add("oscillation_angle", double_t, 0,
-            "How far in radian the robot must rotate to be considered not to be oscillating", 3.14, 0, 6.28)
+            "How far in radians the robot must rotate to be considered not to be oscillating", math.pi, 0, 2*math.pi)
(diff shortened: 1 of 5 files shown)


mbf_abstract_nav/src/mbf_abstract_nav/__init__.py
```diff
@@ -38,3 +38,5 @@ def add_mbf_abstract_nav_params(gen):
             "How long in seconds to allow for oscillation before executing recovery behaviors", 0.0, 0, 60)
     gen.add("oscillation_distance", double_t, 0,
             "How far in meters the robot must move to be considered not to be oscillating", 0.5, 0, 10)
+    gen.add("oscillation_angle", double_t, 0,
+            "How far in radian the robot must rotate to be considered not to be oscillating", 3.14, 0, 6.28)
```
siferati, line 42:
```suggestion
            "How far in radians the robot must rotate to be considered not to be oscillating", math.pi, 0, 2*math.pi)
```
Probably need to include `math` at the top for this to work I guess?

siferati (approved):
@corot So it seems like mbf already has dynamic config for the oscillation parameters, but we're not using them for the exe_path action? Wouldn't it be better to re-use them instead of reading them from parameter server? :thinking: 

https://github.com/renan028/move_base_flex/blob/567f2c8eedc58211fba889ee6f604adfdd9dc06f/mbf_abstract_nav/src/controller_action.cpp#L118-L123

mbf_abstract_nav/src/mbf_abstract_nav/__init__.py
```diff
@@ -38,3 +38,5 @@ def add_mbf_abstract_nav_params(gen):
             "How long in seconds to allow for oscillation before executing recovery behaviors", 0.0, 0, 60)
     gen.add("oscillation_distance", double_t, 0,
             "How far in meters the robot must move to be considered not to be oscillating", 0.5, 0, 10)
+    gen.add("oscillation_angle", double_t, 0,
+            "How far in radians the robot must rotate to be considered not to be oscillating", math.pi, 0, 2*math.pi)
```
siferati: You need to add `import math` at the top... `NameError: name 'math' is not defined`

---

@corot Actually I don't know if this is better or worse than just typing `3.14` and `6.28` instead :joy: It kinda looks ugly on rqt :facepalm: 

image

siferati (changes requested):

mbf_abstract_nav/src/mbf_abstract_nav/__init__.py
knorth55: I reverted the commit because the values looks ugly...

siferati (approved):

mbf_abstract_nav/src/move_base_action.cpp
```diff
@@ -150,6 +152,7 @@ void MoveBaseAction::start(GoalHandle &goal_handle)
     return;
   }
   goal_pose_ = goal.target_pose;
+  last_oscillation_pose_ = robot_pose_;
```
corot, line 155: @knorth55 , is this needed?  if so, was it working before?

corot (approved):

corot:
> @corot So it seems like mbf already has dynamic config for the oscillation parameters, but we're not using them for the exe_path action? Wouldn't it be better to re-use them instead of reading them from parameter server? 🤔
> 
> https://github.com/renan028/move_base_flex/blob/567f2c8eedc58211fba889ee6f604adfdd9dc06f/mbf_abstract_nav/src/controller_action.cpp#L118-L123

right,,,,  https://github.com/magazino/move_base_flex/pull/338

PR #337, opened 2023-11-21
Create 2 topics for current goal: controller and planner

corot:
Also, I cannot see why the controller's topic is created in the abstract_navigation_server and passed all along, so I move to the controller_action (same as planner).

I did the clang formatting required to pass CI in a 2nd commit, to make reviewing easier

### DIFF
--- a/mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
+++ b/mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
@@ -85,7 +85,6 @@ namespace mbf_abstract_nav
      * @param controller_ptr Pointer to the controller plugin
      * @param robot_info Current robot state
      * @param vel_pub Velocity publisher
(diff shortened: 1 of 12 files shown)


mbf_abstract_nav/include/mbf_abstract_nav/abstract_controller_execution.h
siferati, line 1: > I cannot see why the controller's topic is created in the abstract_navigation_server and passed all along

I guess to avoid creating a new publisher every time there is a new exe_path action? But we were already creating a new publisher for every new get_path action anyway so... :shrug:

siferati (approved):

RainerKuemmerle:
Do we by this end up with more sockets in a WAIT state with multiple ExePath calls?

corot:
> Do we by this end up with more sockets in a WAIT state with multiple ExePath calls?

No, no,,,  ControllerAction is created only once at start. Executions are the objects created with each ExePath call

PR #338, opened 2023-11-24
ControllerAction: use dynamic params on oscillation detection

corot:
Same as MoveBaseAction does

I have renamed AbstractActionBase::reconfigureAll as AbstractActionBase::reconfigure, so any action that needs dynamic parameters can just override it (as ControllerAction does now)

### DIFF
--- a/mbf_abstract_nav/include/mbf_abstract_nav/controller_action.h
+++ b/mbf_abstract_nav/include/mbf_abstract_nav/controller_action.h
@@ -90,7 +90,7 @@ protected:
   //! Publish the current goal pose (the last pose of the path we are following)
   ros::Publisher goal_pub_;
 
-  //! timeout after a oscillation is detected
+  //! timeout after an oscillation is detected
   ros::Duration oscillation_timeout_;
 
   //! minimal move distance to not detect an oscillation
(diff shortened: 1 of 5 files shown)


mbf_abstract_nav/include/mbf_abstract_nav/controller_action.h
```diff
@@ -95,6 +97,14 @@ class ControllerAction :
   geometry_msgs::PoseStamped robot_pose_; ///< Current robot pose
   geometry_msgs::PoseStamped goal_pose_;  ///< Current goal pose
 
+  //! timeout after a oscillation is detected
```
siferati:
```suggestion
  //! timeout after an oscillation is detected
```
F: typo

mbf_abstract_nav/src/controller_action.cpp
```diff
@@ -115,16 +124,6 @@ void ControllerAction::runImpl(GoalHandle &goal_handle, AbstractControllerExecut
 
   ros::NodeHandle private_nh("~");
 
-  double oscillation_timeout_tmp;
-  private_nh.param("oscillation_timeout", oscillation_timeout_tmp, 0.0);
-  ros::Duration oscillation_timeout(oscillation_timeout_tmp);
-
-  double oscillation_distance;
-  private_nh.param("oscillation_distance", oscillation_distance, 0.03);
```
siferati, line 127: There is a slight change of behavior -- before the param was defaulting to `0.03`, but now it defaults to `0.5`.
`0.5` seems like a better value so I think is ok though

siferati (approved):

mbf_abstract_nav/src/controller_action.cpp
corot: good point

PR #340, opened 2023-12-11
Implement service to force update the costmaps

siferati:
## Description

Added a new service `force_update_costmaps` to force an update to happen NOW.

**Bonus:** fixed description of `clear_costmaps` service -- I guess it was a mistake from copy-pasting.

## Testing

```bash
> rosservice call /move_base_flex/force_update_costmaps "{}"
```

## Demo

Note that in the video I set the update frequency to 0, so in order for rviz to reflect the change I need to re-toggle the subscribers.
But you can see how the footprint immediately updates to the new position.

https://github.com/magazino/move_base_flex/assets/15384781/1bbfb213-78ad-4349-a14c-78823f1f96e6

### DIFF
--- a/mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
+++ b/mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
@@ -175,13 +175,12 @@ CostmapNavigationServer::CostmapNavigationServer(const TFPtr& tf_listener_ptr)
       private_nh_.advertiseService("check_pose_cost", &CostmapNavigationServer::callServiceCheckPoseCost, this);
   check_path_cost_srv_ =
       private_nh_.advertiseService("check_path_cost", &CostmapNavigationServer::callServiceCheckPathCost, this);
-  clear_costmaps_srv_ =
-      private_nh_.advertiseService("clear_costmaps", &CostmapNavigationServer::callServiceClearCostmaps, this);
   find_valid_pose_srv_ =
       private_nh_.advertiseService("find_valid_pose", &CostmapNavigationServer::callServiceFindValidPose, this);
-
-  force_update_costmaps_srv_ =
-      private_nh_.advertiseService("force_update_costmaps", &CostmapNavigationServer::callForceUpdateCostmaps, this);
+  update_costmaps_srv_ =
+      private_nh_.advertiseService("update_costmaps", &CostmapNavigationServer::callServiceUpdateCostmaps, this);
+  clear_costmaps_srv_ =
+      private_nh_.advertiseService("clear_costmaps", &CostmapNavigationServer::callServiceClearCostmaps, this);
 
   // dynamic reconfigure server for mbf_costmap_nav configuration; also include abstract server parameters
   dsrv_costmap_ = boost::make_shared<dynamic_reconfigure::Server<mbf_costmap_nav::MoveBaseFlexConfig> >(private_nh_);
@@ -821,10 +820,10 @@ bool CostmapNavigationServer::callServiceClearCostmaps(std_srvs::Empty::Request&
   return true;
 }
 
-bool CostmapNavigationServer::callForceUpdateCostmaps(std_srvs::Empty::Request& request,
-                                                      std_srvs::Empty::Response& response)
+bool CostmapNavigationServer::callServiceUpdateCostmaps(std_srvs::Empty::Request& request,
+                                                        std_srvs::Empty::Response& response)
 {
-  // force update both costmaps
+  // update both costmaps
   for (auto costmap : { local_costmap_ptr_, global_costmap_ptr_ })
   {
     costmap->checkActivate();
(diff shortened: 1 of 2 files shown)


mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
```diff
@@ -817,6 +820,20 @@ bool CostmapNavigationServer::callServiceClearCostmaps(std_srvs::Empty::Request&
   return true;
 }
 
+bool CostmapNavigationServer::callForceUpdateCostmaps(std_srvs::Empty::Request& request,
+                                                      std_srvs::Empty::Response& response)
+{
+  // force update both costmaps
+  for (auto costmap : { local_costmap_ptr_, global_costmap_ptr_ })
```
ChristofDubs, line 827: Suggestion: I guess an improvement would be if one can specify in the request which costmap to update? I particularly think force updating the local costmap isn't necessary in most applications because its update rate is usually fairly high.

ChristofDubs (approved):

mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
siferati: Yes I thought about that but I think it would make the service too verbose.
In the end I think this service is more similar to the `clear_costmaps` service rather than the `check_point_cost` service so I decided to keep it as Empty for ease of use

ChristofDubs: Hmm all the services though in [mbf_msgs/srv](https://github.com/magazino/move_base_flex/tree/master/mbf_msgs/srv) have a `uint8 costmap` field; so arguably `clear_costmaps` is the odd one out...

ChristofDubs: I also think being able to clear just a single costmap would be a reasonable requirement, or rather there seems no reason to enforce having to clear both costmaps?

corot: All new services added to MBF force choosing the costmap because doing on both can give different results
That said, I agree it makes sense to allow choosing the cm on both clear and update
:thinking: 
But changing old good clear_costmaps API.... 
:thinking: :thinking: :thinking: 
not sure, but if we have no usecase for updating only one, I think we can let this as it is, instead of breaking the symmetry with clear_costmaps

mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
```diff
@@ -180,6 +180,9 @@ CostmapNavigationServer::CostmapNavigationServer(const TFPtr& tf_listener_ptr)
   find_valid_pose_srv_ =
       private_nh_.advertiseService("find_valid_pose", &CostmapNavigationServer::callServiceFindValidPose, this);
 
+  force_update_costmaps_srv_ =
+      private_nh_.advertiseService("force_update_costmaps", &CostmapNavigationServer::callForceUpdateCostmaps, this);
```
corot: S: `update_costmaps`;  we are not forcing anything,,,  :thinking:

corot: and callForceUpdateCostmaps -> callServiceUpdateCostmaps (same as the others)

mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
```diff
@@ -180,6 +180,9 @@ CostmapNavigationServer::CostmapNavigationServer(const TFPtr& tf_listener_ptr)
   find_valid_pose_srv_ =
       private_nh_.advertiseService("find_valid_pose", &CostmapNavigationServer::callServiceFindValidPose, this);
 
+  force_update_costmaps_srv_ =
```
corot: S: move clear_costmaps after this one, as a more logic grouping

corot (changes requested):
remove the `force_`

mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
siferati: Done

mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
siferati: oops I didn't notice all of the callbacks had "service" in the name :facepalm:

mbf_costmap_nav/src/mbf_costmap_nav/costmap_navigation_server.cpp
siferati: Upon further discussion we decided to keep the current implementation of both `clear_costmaps` and `update_costmaps` services being Empty

corot (approved):

PR #341, opened 2023-12-15
change plugin loader to load simple_core plugins

mrvsht:
This change is needed to load simple_core plugins.
Otherwise abstract_core plugins are loaded, which results in an incorrect start of mbf.


Cakem1x:
afaik, this is outdated. @mrvsht  please re-open in case I am wrong :)

PR #343, opened 2024-04-26
Fix 3 related issues on move_base action recovering

corot:
- `move_base` action get's stuck if recovery behavior patience is exceeded
- If a recovery behavior fails, calling the next immediatly breaks simple action client
- If the last recovery behavior fails, we return that as move_base error instead of the error that triggered recovering

I was debugging the first, but realized the 2 other on the process

The solution for the first is to treat patience exceeded the same way as behavior failure. Then I realized the other 2 errors.
The solution for the 2nd is [rather hackish](https://github.com/magazino/move_base_flex/compare/js/fix_move_base_action?expand=1#diff-6b8a12259c3d11b5f68c080c8a6d7a690aecea38015c66366f1d559b16bd4acbR481-R482), but a proper solution would require to implement the move_base action, for which I don't have time. In any case the previous code was incorrect.

### DIFF
--- a/mbf_abstract_nav/src/move_base_action.cpp
+++ b/mbf_abstract_nav/src/move_base_action.cpp
@@ -125,8 +125,6 @@ void MoveBaseAction::start(GoalHandle &goal_handle)
 
   const mbf_msgs::MoveBaseGoal& goal = *goal_handle.getGoal();
 
-  mbf_msgs::MoveBaseResult move_base_result;
-
   get_path_goal_.target_pose = goal.target_pose;
   get_path_goal_.use_start_pose = false; // use the robot pose
   get_path_goal_.planner = goal.planner;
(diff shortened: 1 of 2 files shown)


mbf_abstract_nav/src/move_base_action.cpp
```diff
@@ -454,45 +449,41 @@ void MoveBaseAction::actionRecoveryDone(
...
+        // we run out of recovery behaviors; concede defeat
+        action_state_ = FAILED;
+        ROS_DEBUG_STREAM_NAMED("move_base", "All recovery behaviors failed. Aborting move_base action");
+        goal_handle_.setAborted(move_base_result_, "All recovery behaviors failed.");
+        break;
       }
-      break;
+      // recovery failed but we still have behaviors; we cannot call the next here (simple action client
+      // will complain about transition to ACTIVE from DONE), so we try replanning again (wrong, but works)
     case actionlib::SimpleClientGoalState::SUCCEEDED:
```
siferati:
```suggestion
      // recovery failed but we still have behaviors; we cannot call the next here
      // (simple action client will complain about transition to ACTIVE from DONE)
      // so we try replanning again (wrong, but works)
      [[fallthrough]];
    case actionlib::SimpleClientGoalState::SUCCEEDED:
```
S: add [`[[fallthrough]]`](https://en.cppreference.com/w/cpp/language/attributes/fallthrough) attribute to make it clear what the comment is referring to.
I was a bit confused at first until I noticed it was referring to the lack of a `break` statement 😅
I think this requires C++17 though, so if you don't want to change `CMakeLists.txt` then maybe add it commented out

S: better text line breaks

siferati (approved):

Issue #323, opened 2023-07-21
MBF will block the main thread for an arbitrary time when canceling an execution


corot:
Because [we join its thread after cancelling an execution](https://github.com/magazino/move_base_flex/blob/master/mbf_abstract_nav/include/mbf_abstract_nav/abstract_action_base.hpp#L147). This is dangerous because we relay on plugins cancel implementation to finish properly. And meanwhile we won't have callbacks, so for example a controller slowing down on cancel won't get an updated velocity from MBF.

**UPDATE** After #345, this problem won't happen if the old and new goals have the same controller. In that case, we will preempt the previous goal instead of canceling it.

Nisarg236:
Hey @corot @siferati @renan028 , I was working on this issue and this implementation works: https://github.com/naturerobots/move_base_flex/pull/359 but I wanted to know if you thought about something similar and there is any side effects of doing so.

PR #345, opened 2024-08-21
Update running controller goal also when we are preempting it

corot:
This PR partially addresses issue #323, for the case of preempting a controller execution with the same controller.
The problem is still there if the old and new goals have different controller.

I re-state here the problem and the solution
1. force_stop_on_cancel is set to false (we let the controller to smoothly stop when canceled)
2. goal running
3. we cancel it;  the controller slows down gently
4. before stopping, a new controller goal arrives
5. we [cancel the previous execution and join the thread](https://github.com/magazino/move_base_flex/blob/master/mbf_abstract_nav/include/mbf_abstract_nav/abstract_action_base.hpp#L144-L149), waiting for the robot to stop
6. we keep calling the canceled controller to get slowing down speeds
7. BUT as we have blocked the main thread, the velocity we pass to it (provided by /odom topic) won't change
8. If the controller happens to relay on this velocity to control it's slowing down and stopping, we can deadlock

So instead of canceling and joining, with this PR, things change from point 5

5. we reuse the same slot and change the path of the current execution
6. it's not canceled anymore, so it executes the new path

### DIFF
--- a/mbf_abstract_nav/include/mbf_abstract_nav/abstract_action_base.hpp
+++ b/mbf_abstract_nav/include/mbf_abstract_nav/abstract_action_base.hpp
@@ -143,7 +143,7 @@ public:
         // if there is already a plugin running on the same slot, cancel it
         slot_it->second.execution->cancel();
 
-        // TODO + WARNING: this will block the main thread for an arbitrary time during which we won't execute callbacks
+        // WARNING: this will block the main thread for an arbitrary time during which we won't execute callbacks
         if (slot_it->second.thread_ptr->joinable()) {
           slot_it->second.thread_ptr->join();
         }
--- a/mbf_abstract_nav/src/controller_action.cpp
+++ b/mbf_abstract_nav/src/controller_action.cpp
@@ -81,9 +81,10 @@ void ControllerAction::start(
   if(slot_it != concurrency_slots_.end() && slot_it->second.in_use)
   {
     boost::lock_guard<boost::mutex> goal_guard(goal_mtx_);
+    const auto slot_status = slot_it->second.goal_handle.getGoalStatus().status;
     if ((slot_it->second.execution->getName() == goal_handle.getGoal()->controller ||
          goal_handle.getGoal()->controller.empty()) &&
-        slot_it->second.goal_handle.getGoalStatus().status == actionlib_msgs::GoalStatus::ACTIVE)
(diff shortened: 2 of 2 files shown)


mbf_abstract_nav/src/controller_action.cpp
```diff
@@ -81,9 +81,10 @@ void ControllerAction::start(
   if(slot_it != concurrency_slots_.end() && slot_it->second.in_use)
   {
     boost::lock_guard<boost::mutex> goal_guard(goal_mtx_);
+    const auto slot_status = slot_it->second.goal_handle.getGoalStatus().status;
     if ((slot_it->second.execution->getName() == goal_handle.getGoal()->controller ||
          goal_handle.getGoal()->controller.empty()) &&
-        slot_it->second.goal_handle.getGoalStatus().status == actionlib_msgs::GoalStatus::ACTIVE)
+        (slot_status == actionlib_msgs::GoalStatus::ACTIVE || slot_status == actionlib_msgs::GoalStatus::PREEMPTING))
```
corot, line 87: equivalent to [isActive method on SimpleActionServer](https://docs.ros.org/en/indigo/api/actionlib/html/simple__action__server__imp_8h_source.html#l00228)

mbf_abstract_nav/include/mbf_abstract_nav/abstract_action_base.hpp
```diff
@@ -143,7 +143,7 @@ class AbstractActionBase
         // if there is already a plugin running on the same slot, cancel it
         slot_it->second.execution->cancel();
 
-        // TODO + WARNING: this will block the main thread for an arbitrary time during which we won't execute callbacks
+        // WARNING: this will block the main thread for an arbitrary time during which we won't execute callbacks
```
corot, line 146: Still true, e.g. if the old and new goals have different controller.

siferati (approved):
This PR only fixes the issue if the new goal has the same controller as the currently cancelling goal right? Maybe better to leave the issue open then even after merging this PR.

Glad to see the graceful cancel bug at least partially addressed though! 🥳

corot:
> This PR only fixes the issue if the new goal has the same controller as the currently cancelling goal right? Maybe better to leave the issue open then even after merging this PR.

correct
yes, it makes sense to keep the issue as open
thanks for the review!

Issue #346, opened 2024-10-16
Finalize ROS2 Migration


Cakem1x:
The [humble branch](https://github.com/naturerobots/move_base_flex/tree/humble) currently tracks the latest move base flex version for ROS2.

I want to tackle the following post migration ToDos:
* [x] update default branch: from `master` to `humble`
* [x] branch protection rule for `humble`
* [x] update readmes (ROS2 readme (`humble`), ROS1 readme (`master`))
  * add logos for companies
  * add/update notice about and link to ROS1 and ROS2 branch
  * add news/announcements section, include info about which parts are migrated and which are not.

Cakem1x:
Before changing the default branch to `humble`, I'd like to get confirmation from other maintainers on whether that will break anything critical (@spuetz @corot ).
The rationale for this change is that it makes the landing of new people on the [repo's main page](https://github.com/naturerobots/move_base_flex) nicer, assuming most of them are interested in using ROS2.

corot:
> Before changing the default branch to `humble`, I'd like to get confirmation from other maintainers on whether that will break anything critical (@spuetz @corot ). The rationale for this change is that it makes the landing of new people on the [repo's main page](https://github.com/naturerobots/move_base_flex) nicer, assuming most of them are interested in using ROS2.

I think the default branch should always be `master`. But if most new developments will be on ROS 2, it makes sense to 
- rename master as noetic
- rename humble as master
But as migration is still wip (costmap server missing), we need to wait to do that

Besides that, I think there are changes in jazzy that break compatibility with costmaps, so future master should work for jazzy, not humble

amock:
@corot, @Cakem1x

As we've got a few side projects going on that use different map representations (like mesh maps, OpenVDB), it got me thinking: do we really need to include a costmap_2d implementation in this repo?

The more I thought about it, the more I liked the idea of **not** providing any specific map representation in this repo at all.
With solid docs, I actually think it makes it easier to understand what MBF is really about. It's not limited to a particular map representation.

Otherwise, people might confuse MBF with another ROS 2 framework for navigation on 2D costmaps.
One thing that actually gets hard is testing. However, I really think we can find a good solution to that.

What do you think of that idea?

Cakem1x:
I am good with keeping implementations for different concrete map representations in **separate repos**. Especially if they come with a lot of additional dependencies. Having an overview of where to find map specific implementations would be very important, though.

[...]

PR #350, opened 2025-04-14
Remove Costmap 2D

amock:
Removed costmap 2d as discussed in #346


Rayman:
For me its a bit strange that the costmap 2d is not included in the core mbf repo. This means mbf is not a plug and play replacement for navigation2. This means adoption will be slower.

You can also choose to just not release these packages yet, and wait until they are migrated?

(I'm not a maintainer, I've only used mbf during ROS 1)

amock:
Actually, I don't think MBF itself can be a plug-and-play replacement for Nav2; only a 2d costmap-based implementation of MBF can be. If you want, you can still create a separate repository with such an implementation, which could serve as a plug-and-play replacement for Nav2.

I believe this step simply makes MBF truly and more obviously independent of any specific map representation. That said, I'm not sure if this is the best approach, which is why I started the discussion in issue #346.

Cakem1x (approved):
lgtm. No need to keep the old ROS1 code around here.

PR #356, opened 2025-11-06
Modernize README

amock:
I believe some parts of the README for the ROS 2 version of MBF feel a bit outdated. For instance, many newer ROS users are likely unfamiliar with the original move_base. I've therefore revised the README to better highlight MBF's current purpose and core features, focusing on its role and key design aspects rather than its historical context.

What do you think? I'm open to any feedback or ideas for improvement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,37 +1,32 @@
-This is the active ROS2 branch of this repository. If your are looking for the ROS1 version, checkout the [noetic](https://github.com/naturerobots/move_base_flex/tree/noetic) or [master](https://github.com/naturerobots/move_base_flex/tree/master) branches.
-
 [![Jazzy CI](https://github.com/naturerobots/move_base_flex/actions/workflows/jazzy.yaml/badge.svg)](https://github.com/naturerobots/move_base_flex/actions/workflows/jazzy.yaml)
 [![Humble CI](https://github.com/naturerobots/move_base_flex/actions/workflows/humble.yaml/badge.svg)](https://github.com/naturerobots/move_base_flex/actions/workflows/humble.yaml)
 
-# Move Base Flex: A Highly Flexible Navigation Framework:
+# Move Base Flex: A Highly Flexible Navigation Framework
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -59,3 +68,11 @@ Move Base Flex was initially developed at Magazino.
 
 ### [<img width="25" height="25" src="doc/images/logos/nature_robots_icon.jpg"> Nature Robots](https://naturerobots.com/)
 The latest version (ROS2) is developed and maintained by Nature Robots.
+
+## Further Resources
```
Cakem1x, line 69: most of the resources are ROS1 only. Except:
- https://github.com/naturerobots/mesh_navigation
- https://github.com/amock/mbf_deliberation

I think the resources that only work with ROS1 could be dropped from this list.

README.md
```diff
@@ -1,37 +1,32 @@
...
+  
+  Multiple planners, controllers, and recovery behaviors can be loaded simultaneously, selected at runtime, or executed in parallel using different concurrency slots.
 
-These two packages not migrated:
-- mbf_costmap_core & mbf_costmap_nav (for navigation components that utilize a 2D costmap). Nav2, which hosts the 2D costmap equivalent to the one from ROS1, and ROS1's move_base are quite different, so interfaces do not easily fit anymore. This makes migration hard. PRs are welcome for this. However, we might integrate another 2D grid map planning module soon.
+* **Rich Feedback and Diagnostics**
+  
+  All actions expose detailed feedback, results, and error codes, providing transparent runtime information for better debugging and system supervision.
 
-Note that [mesh_navigation](https://github.com/naturerobots/mesh_navigation) is also available for ROS2, now. It provides navigation components that utilize 3D mesh maps.
```
Cakem1x, line 34: I think keeping a high level overview over what has changed is useful.
I agree that this does not need to part of the readme, where the focus should be the current state, not the history.
I also we could lose some of the text I wrote regarding "16.10.2024 First ROS2 Version of Move Base Flex" :)

Can we move the info to another file or at the end of the readme?

Or do you think that this kind of info should not be present at all here?

Cakem1x (commented):
Good idea to go over the readme. I like the changes, I think they are a large improvement.

README.md
amock: I also thought about keeping the news somewhere. It would be fine to move it to the end of the file. Or maybe put it to the docs?  https://naturerobots.github.io/mbf_docs/ . But those are still ROS 1, I think (?).

README.md
amock: Ah yes, I will remove them with the next commit :+1:

README.md
amock: with the latest commit I readded the news to the end of the readme

Cakem1x (approved):
Thanks for iterating, looks good to me!

PR #357, opened 2025-12-19
Add simple retry logic if the exe_path action failes

JustusBraun:
With these changes RViz retries a goal up to 2 times if the ExePath action returns the ROBOT_STUCK failure code.


amock (approved):

amock:
This is definitely an improvement, as it results in higher success rates for first-time users. For development, however, I think it's beneficial not to retry automatically, because in that context you sometimes want things to fail. So maybe in future PRs we could add an additional field to either specify the exact number of retries or simply toggle this functionality with a radio button.

PR #359, opened 2025-12-22
fix: main thread is blocked when canceling an execution

Nisarg236:
[Issue](https://github.com/naturerobots/move_base_flex/issues/323)
Before, the join was called in the main thread. So when cancel is running, it will block other callbacks like odom callback from running. When cancel is running, the controller relies on odom callback to check if the robot is stopped and then it returns cancel. But as odom is not updated, the cancel keeps running.

After this change, it moves the block that runs cancel and updates the slot in a function and this function is called in a separate thread. So it does not block the main thread. I tested this by sending a goal with a different controller when cancel from one controller is running, and it seems to work. The next goal is executed after the first one is cancelled and the callbacks keep running in parallel.


Nisarg236:
will open from a different branch later

PR #362, opened 2026-02-11
Fix Test Abstract Nav

### LINKED ISSUE #361: Failed Test on Jankins Build Farm
https://build.ros2.org/job/Hdev__move_base_flex__ubuntu_jammy_amd64/1/

```bash
[ RUN      ] AbstractPlannerExecutionFixture.cancel
/tmp/ws/src/move_base_flex/mbf_abstract_nav/test/abstract_planner_execution.cpp:119: Failure
Expected equality of these values:
  planner_execution_ptr_->waitForStateUpdate(std::chrono::seconds(1))
    Which is: 4-byte object <01-00 00-00>
  std::cv_status::no_timeout
    Which is: 4-byte object <00-00 00-00>
[  FAILED  ] AbstractPlannerExecutionFixture.cancel (1009 ms)
```

We should reproduce this failing test soon. And then find a fix. Also: Why are the Github workflows not failing?

amock:
This PR removes the asserts on the return value of `planner_execution_ptr_->waitForStateUpdate` in the tests for `AbstractPlannerExecution`. Here’s why I think this is acceptable:

* MBF core itself (`planner_action.cpp`) does not fail if this return value times out.
* The same approach is already used in the `AbstractControllerExecution` tests.

We also reintroduced workflows for the new ROS 2 `main` branch.

This should fix #361. However, we should still review the code to see whether we can make it safer with respect to threading, perhaps by introducing some kind of promise/future mechanism. I will create a new issue for that.


mbf_abstract_nav/test/abstract_action_base.cpp
GeorgJohn, line 1: These changes fix another test failure.

GeorgJohn (approved):
Nice work cleaning these up! Tests are running now on all platforms without failing for me. However, the method `waitForStateUpdate(...)` should return correct and reliable values in the future.

mbf_abstract_nav/test/abstract_action_base.cpp
amock: I forgot to mention: There was an additional segfault because the ROS context was initialized outside the Google Test scope. I assume this is what caused the segfault. I changed it to follow the style used in the other tests.

amock:
I added an issue for that: #363

PR #380, opened 2026-03-17
rviz_mbf_plugins: Select planners and controllers from dropdown menu

amock:
For the RViz MBFActionPanel, I added functionality to select controllers and planners via a dropdown list. The list is filled through a parameter service call to the MBF instance.

One issue remains: the node name is currently inferred by splitting the action topic. This approach breaks if the action is remapped. However, this is acceptable for now, as the panel is primarily intended as a debugging/development tool.

Minor fix: ROSTopicProperty -> ROSActionProperty for all the actions


JustusBraun (commented):
Works with latest `mesh_navigation_tutorials` on ROS Jazzy (Ubuntu 24)

PR #383, opened 2026-06-17
Static -> Dynamic Pointer Casts

amock:
This PR changes static pointer casts to dynamic ones, as this allows us to make safer checks on whether the cast from the base to the derived class was actually successful. With this change, the return value starts to make sense.


GeorgJohn (approved):
LGTM

PR #385, opened 2026-07-03
Fix RobotInformation Tf-Lookup Latency Slows Control Loop Down

JustusBraun:
I was using MeshNav on a real robot when I encountered a problem where the move_base_flex controller execution would print warnings, that calculating the control signal took too long, despite the calculation time measured inside the controller plugin being well below the loop cycle length. I eventually found out that the whole control loop was slowed down by the robot pose lookup: https://github.com/naturerobots/move_base_flex/blob/43173aa9b7b7cfeab9fb845448642f7fce2e33d1/mbf_abstract_nav/src/abstract_controller_execution.cpp#L480-L491

The `getRobotPose` method did a  TF-Lookup between the map and base frames at time `node->get_clock()->now()`, which then blocks until the transform is available. When using the now time TF2 waits until a newer transform is available and then interpolates to the requested time. In my case this took around 150 ms.
https://github.com/naturerobots/move_base_flex/blob/43173aa9b7b7cfeab9fb845448642f7fce2e33d1/mbf_utility/src/robot_information.cpp#L59-L75

So I tried using time point zero for the lookup, which requests the latest available transform from TF2. This did restore the 20 Hz control loop, however, this method returns the transform for the latest time, where all links in the transform chain have a transform available. So with a typical robot setup with a transform chain `base_link` -> `odom` -> `map`, the total transform is only updated with the frequency of the slowest transform. In my case this is the `odom` -> `map` transform, since it is only updated with the LiDAR frequency, which is below the 20 Hz control rate. This can lead to degraded control performance since the control algorithm uses an outdated pose estimate.

For the control loop I would expect to always use the latest estimate of the robot position inside the map. This would be using the latest available `odom` -> `map` transform chained with the latest `base_link` -> `odom` transform, since the odometry updates at a much higher frequency. This method has the additional advantage that no interpolation is performed at any time.

Because TF2 cannot do this directly (at least to my knowledge) I implemented this as a two part lookup. I had to add a new parameter for the odom frame to the navigation server. With this change everything worked as expected.

### DIFF
--- a/mbf_abstract_nav/include/mbf_abstract_nav/abstract_navigation_server.h
+++ b/mbf_abstract_nav/include/mbf_abstract_nav/abstract_navigation_server.h
@@ -368,6 +368,9 @@ protected:
   //! the robot frame, to get the current robot pose in the global_frame_
   std::string robot_frame_;
 
+  //! the odometry frame, in which the robot is moving
+  std::string odom_frame_;
+
   //! the global frame, in which the robot is moving
   std::string global_frame_;
 
--- a/mbf_abstract_nav/src/abstract_navigation_server.cpp
+++ b/mbf_abstract_nav/src/abstract_navigation_server.cpp
@@ -69,6 +69,7 @@ AbstractNavigationServer::AbstractNavigationServer(
     node)
 {
   node_->declare_parameter<std::string>("global_frame", "map");
+  node_->declare_parameter<std::string>("odom_frame", "odom");
   node_->declare_parameter<std::string>("robot_frame", "base_link");
   node_->declare_parameter<double>("tf_timeout", 3.0);
   node_->declare_parameter<std::string>("odom_topic", "~/odom");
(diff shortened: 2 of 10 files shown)


amock:
I also think this is a very important part to optimize. It's crucial a control step doesn't waste time especially if it's not required. The following is a small visual example of the general transform-chain problem:

`T_o_b_0`: Transform from base to odom at timestamp 0. Adding a usual timeseries of transforms:

| t             | 0 | 1 | 2 | 3 |
|:------------|:----------|:------------|:----------|:--------|
| `T_m_o` |  | `T_m_o_1` | | | 
| `T_o_b` | `T_o_b_0` | | `T_o_b_2` | `T_o_b_3` |

So for example at timestamp 2 we are searching for `T_m_b_2`.  Which is

```
T_m_b_2 = T_m_o_2 * T_o_b_2
```

`T_m_o_2` is not given, so we take the latest estimate, which in this case is `T_m_o_1`. When searching for `T_m_b_3`, we would also use `T_m_o_1`. I think this means we assume the velocity of the drift between odom and map to be zero. Still, it seems to be way better than before. I'll check your changes out soon.

amock (approved):

PR #386, opened 2026-07-07
Rviz Plugin Improvements

amock:
This PR adds some stability improvements to the RViz MBF plugin. It's still not perfect, but more usable than before. Since it is not part of the MBF interfaces, it also doesn't change any interfaces and is completely backwards compatible.


GeorgJohn (approved):
I don't see any reason why not to merge. LGTM

PR #387, opened 2026-07-07
ROS 2 Lyrical Compatibility 

amock:
- removed `ament_target_dependencies`
- new cmake-target-based CMakeLists. Kept also old style "_TARGETS" variable for backwards compatibility
- compiled with humble, jazzy, lyrical
- tested with MeshNav under jazzy

### DIFF
--- a/.github/workflows/humble.yaml
+++ b/.github/workflows/humble.yaml
@@ -12,7 +12,7 @@ on:
 
 jobs:
   humble_build_and_test:
-    uses: naturerobots/github_automation_public/.github/workflows/ros_ci.yaml@main
+    uses: naturerobots/github_automation_public/.github/workflows/ros_ci.yaml@cc7fc175054179167e8a676c0e5e11bbbfe60d4e
     secrets: inherit
     with:
       ros_distro: humble
--- a/.github/workflows/jazzy.yaml
+++ b/.github/workflows/jazzy.yaml
@@ -12,7 +12,7 @@ on:
 
 jobs:
   jazzy_build_and_test:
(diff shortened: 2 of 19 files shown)


amock:
Before merging, we have to deal with
- the conditional qt6 or qt5 version in the package.xml of rviz_mbf_plugins
- adding a lyrical workflow

JustusBraun:
For the conditional dependencies it should be possible to just use the condition feature of the package.xml spec:
https://www.ros.org/reps/rep-0149.html#dependency-tags

Example:
```xml
<depend condition="$ROS_DISTRO == jazzy">qt5<depend>
<depend condition="$ROS_DISTRO == lyrical">qt6<depend>
```

amock:
My current plan is to avoid those conditionals by using a more generic way of include the qt deps. Without the version number. I think rviz_common does the same

amock:
> For the conditional dependencies it should be possible to just use the condition feature of the package.xml spec: https://www.ros.org/reps/rep-0149.html#dependency-tags
> 
> Example:
> 
> ```
> <depend condition="$ROS_DISTRO == jazzy">qt5<depend>
> <depend condition="$ROS_DISTRO == lyrical">qt6<depend>
> ```

I was not able to came around some conditionals in the package.xml as you suggested, as rosdep could not resolve e.g. the key `qt-base-dev` on humble. For the newer versions of ROS however it's working. So when we drop the support for humble in the future, we can clean this package.xml up again

GeorgJohn (approved):
LGTM

